PR #18, opened 2016-02-23
Move null checks to builder for compatibility with Strong Mode analyzer.

davidmorgan:


example/test/value_test.dart
```diff
@@ -20,10 +20,14 @@ void main() {
         ..aString = '').aDefaultInt, 7);
     });
 
-    test('throws on null for non-nullable fields', () {
+    test('throws on null for non-nullable fields on build', () {
       expect(() => new Value(), throws);
     });
 
+    test('does not throw on null for non-nullable fields in builder', () {
```
dskloetg: I don't think this test adds value.

dskloetg:
LGTM

example/test/value_test.dart
davidmorgan: > I don't think this test adds value.
 
Removed.

PR #28, opened 2016-06-19
Improve error output on failure to generate.

davidmorgan:
This change is [Reviewable](https://reviewable.io/reviews/google/built_value.dart/28)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,13 @@
 # Changelog
 
+## 0.1.1
+
+- Improve error output on failure to generate.
+
 ## 0.1.0
 
- - Upgrade to source_gen 0.5.0.
- - Breaking change; see example for required changes to build.dart.
+- Upgrade to source_gen 0.5.0.
+- Breaking change; see example for required changes to build.dart.
 
 ## 0.0.6
 
--- a/built_value/pubspec.yaml
+++ b/built_value/pubspec.yaml
@@ -1,5 +1,5 @@
 name: built_value
-version: 0.1.0
+version: 0.1.1
 description: >
   Value types with builders. This library is the runtime dependency.
 authors:
--- a/built_value_generator/lib/built_value_generator.dart
+++ b/built_value_generator/lib/built_value_generator.dart
@@ -47,9 +47,7 @@ class BuiltValueGenerator extends Generator {
     ]);
 
     if (errors.isNotEmpty) {
-      throw new InvalidGenerationSourceError(
-          'Please make changes to use built_value.',
-          todo: errors.join(' '));
+      throw _makeError(errors);
     }
 
     return generateCode(className, fields, builderFields);
@@ -69,7 +67,7 @@ class BuiltValueGenerator extends Generator {
     final name = classElement.displayName;
 
     if (!classElement.isAbstract) {
-      result.add('Make class abstract');
+      result.add('Make class abstract.');
     }
 
     final expectedConstructor = '$name._()';
@@ -100,7 +98,7 @@ class BuiltValueGenerator extends Generator {
   Iterable<String> checkBuilderClass(
       String className, ClassElement classElement) {
     if (classElement == null) {
-      return <String>['Add abstract class ${className}Builder'];
+      return <String>['Add abstract class: ${className}Builder'];
     }
 
     final result = <String>[];
@@ -164,7 +162,7 @@ class BuiltValueGenerator extends Generator {
       final fieldName = field.displayName;
       if (field.getter == null || field.getter.isSynthetic) {
         checkFieldTypes = false;
-        result.add('Make field $fieldName a getter');
+        result.add('Make field $fieldName a getter.');
       }
     }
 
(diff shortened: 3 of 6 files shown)


manuelF:
Review status: 0 of 6 files reviewed at latest revision, 2 unresolved discussions.

---

*[built_value_generator/lib/built_value_generator.dart, line 186 \[r1\]](https://reviewable.io:443/reviews/google/built_value.dart/28#-KKdcq4fzdnEUB7TO8xX:-KKdcq4fzdnEUB7TO8xY:-363378806) ([raw file](https://github.com/google/built_value.dart/blob/5bc6eff8c79d212f0dd20e0860da0179f844f5a9/built_value_generator/lib/built_value_generator.dart#L186)):*
> ```Dart
>       result.add('Make builder have exactly these fields: ' +
>           fieldNames.join(', ') +
>           '.');
> ```

Wouldn't the period make it harder to copy-paste? At least put a space before it so it is simpler. Or write them as lists so it is more evident that they are tokens and not text, e.g: 

X. Make builder have exactly these fields: [foo, bar, baz, buz].

---

*[built_value_generator/lib/built_value_generator.dart, line 333 \[r1\]](https://reviewable.io:443/reviews/google/built_value.dart/28#-KKdeZdtGL5ZN4db3uog:-KKdeZdu6_ya_71czysV:2021556784) ([raw file](https://github.com/google/built_value.dart/blob/5bc6eff8c79d212f0dd20e0860da0179f844f5a9/built_value_generator/lib/built_value_generator.dart#L333)):*
> ```Dart
>     final message = new StringBuffer(
>         'Please make the following changes to use built_value:\n');
>     for (var i = 0; i != todos.length; ++i) {
> ```

i < todos.length is more idiomatic, I believe.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/built_value.dart/28)*

davidmorgan:
Review status: 0 of 6 files reviewed at latest revision, 2 unresolved discussions.

---

*[built_value_generator/lib/built_value_generator.dart, line 186 \[r1\]](https://reviewable.io:443/reviews/google/built_value.dart/28#-KKdcq4fzdnEUB7TO8xX:-KKe0gFr6bVpn40ZgLKp:809535608) ([raw file](https://github.com/google/built_value.dart/blob/5bc6eff8c79d212f0dd20e0860da0179f844f5a9/built_value_generator/lib/built_value_generator.dart#L186)):*
<details><summary><i>Previously, manuelF (Manuel Ferreria) wrote…</i></summary>
> Wouldn't the period make it harder to copy-paste? At least put a space before it so it is simpler. Or write them as lists so it is more evident that they are tokens and not text, e.g: 
> 
> X. Make builder have exactly these fields: [foo, bar, baz, buz].
</details>
Good point, the ones that are code should not have a period. Will revert.

---

*[built_value_generator/lib/built_value_generator.dart, line 333 \[r1\]](https://reviewable.io:443/reviews/google/built_value.dart/28#-KKdeZdtGL5ZN4db3uog:-KKe0l3I5e9Tx9uMjvVK:-409787730) ([raw file](https://github.com/google/built_value.dart/blob/5bc6eff8c79d212f0dd20e0860da0179f844f5a9/built_value_generator/lib/built_value_generator.dart#L333)):*
<details><summary><i>Previously, manuelF (Manuel Ferreria) wrote…</i></summary>
> i < todos.length is more idiomatic, I believe.
</details>
I always write for loops like this :)

It means the loop has a stronger postcondition: it will terminate when i is exactly todos.length, and not when it's todos.length+1, +2, +100...

If you happen to be proving something about what the loop will do, this is much better.

But the reason I like it in practice is that it makes clear that I know i is not going to start bigger than todos.length. Occasionally you don't know whether the loop will execute at all, and if you get into the habit of using !=, you'll have to handle these cases differently. (Which you probably should anyway).

I can change it if you like. Up to you :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/built_value.dart/28)*

manuelF:
Review status: 0 of 6 files reviewed at latest revision, 1 unresolved discussion.

---

*[built_value_generator/lib/built_value_generator.dart, line 333 \[r1\]](https://reviewable.io:443/reviews/google/built_value.dart/28#-KKdeZdtGL5ZN4db3uog:-KKe2pn6DHQa_GBaZQ5e:886869941) ([raw file](https://github.com/google/built_value.dart/blob/5bc6eff8c79d212f0dd20e0860da0179f844f5a9/built_value_generator/lib/built_value_generator.dart#L333)):*
<details><summary><i>Previously, davidmorgan (David Morgan) wrote…</i></summary>
> I always write for loops like this :)
> 
> It means the loop has a stronger postcondition: it will terminate when i is exactly todos.length, and not when it's todos.length+1, +2, +100...
> 
> If you happen to be proving something about what the loop will do, this is much better.
> 
> But the reason I like it in practice is that it makes clear that I know i is not going to start bigger than todos.length. Occasionally you don't know whether the loop will execute at all, and if you get into the habit of using !=, you'll have to handle these cases differently. (Which you probably should anyway).
> 
> I can change it if you like. Up to you :)
</details>
Your reasoning is solid, agree.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/built_value.dart/28)*

manuelF:
:lgtm:

---

Review status: 0 of 6 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/built_value.dart/28#-:-KKe341eiotoe9pVAFPj:1417088721)*

PR #81, opened 2016-12-08
Fix builder getters to be available before set.

### LINKED ISSUE #80: Non-nullable fields are null in rebuild() closure
[I may be doing something wrong, and it's too late for me to try and build a reduced test case, so take this with a grain of salt.]

I have a class like this:

```
abstract class Situation implements Built<Situation, SituationBuilder> {
  int get id;
  SituationState get state;
  // ...
}
```

And an associated builder like this:

```
abstract class SituationBuilder
    implements Builder<Situation, SituationBuilder> {
  @virtual
  int id = _random.nextInt(_largeInteger);
  @virtual
  SituationState state;

  SituationBuilder._();
  factory SituationBuilder() = _$SituationBuilder;
}
```

Note that `state` is not nullable. But when I run `rebuild` on the situation, the provided SituationBuilder instance's `state` field is set to `null`.

I'm on Dart 1.21.0. Problem happens both on Dart VM and in dart2js. The previous version didn't seem to have this issue (although I did recently refactor so I can't be sure).

Here's some debugging flow when `rebuild` is called on a Situation instance.

---

screen shot 2016-12-08 at 00 28 38

---

screen shot 2016-12-08 at 00 29 15

---

screen shot 2016-12-08 at 00 29 57

---

screen shot 2016-12-08 at 00 30 10

---

The last one is the state of the SituationBuilder when it's being passed to the provided `updates()` closure.

davidmorgan:
Renamed "_$writableBuilder" to "_$this", since

 - it's now used for all the setters and getters in the builder, not just setters (which is why it was "writable")
 - a longer name didn't seem to add much, and we might as well keep the generated code a bit shorter

Fixes #80 

---
This change is [Reviewable](https://reviewable.io/reviews/google/built_value.dart/81)

### DIFF
--- a/benchmark/lib/node.g.dart
+++ b/benchmark/lib/node.g.dart
@@ -52,20 +52,20 @@ class NodeBuilder implements Builder<Node, NodeBuilder> {
   Node _$v;
 
   String _label;
-  String get label => _label;
-  set label(String label) => _$writableBuilder._label = label;
+  String get label => _$this._label;
+  set label(String label) => _$this._label = label;
 
   NodeBuilder _left;
-  NodeBuilder get left => _$writableBuilder._left ??= new NodeBuilder();
-  set left(NodeBuilder left) => _$writableBuilder._left = left;
+  NodeBuilder get left => _$this._left ??= new NodeBuilder();
+  set left(NodeBuilder left) => _$this._left = left;
 
   NodeBuilder _right;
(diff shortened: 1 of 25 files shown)


filiph:
Please let me know when this lands on pub. If it's soon enough, I'd love to avoid the whole {clone repository, change pubspec, remember to change back} dance.

davidmorgan:
On pub now. Thanks!

PR #89, opened 2016-12-20
Add StandardJsonPlugin.

### LINKED ISSUE #65: Support "pretty" JSON output
A few people have asked about supporting "standard" JSON, meaning some kind of key->value maps with fieldnames.

It's not yet clearly exactly what would make sense here, because that isn't sufficient to serialize data models that are using interfaces rather than concrete types. Please expand on requirements here if this is what you'd like.

davidmorgan:
Fixes #65

---
This change is [Reviewable](https://reviewable.io/reviews/google/built_value.dart/89)

### DIFF
--- a/built_value/test/standard_json_plugin_test.dart
+++ b/built_value/test/standard_json_plugin_test.dart
@@ -1,0 +1,190 @@
+// Copyright (c) 2016, Google Inc. Please see the AUTHORS file for details.
+// All rights reserved. Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+import 'package:built_collection/built_collection.dart';
+import 'package:built_value/standard_json_plugin.dart';
+import 'package:built_value/serializer.dart';
+import 'package:test/test.dart';
+
+void main() {
+  final serializers = (new Serializers().toBuilder()
+        ..addPlugin(new StandardJsonPlugin())
+        ..addBuilderFactory(
+            const FullType(BuiltList, const [const FullType(int)]),
+            () => new ListBuilder<int>())
+        ..addBuilderFactory(
+            const FullType(BuiltList, const [
+              const FullType(BuiltList, const [const FullType(int)])
+            ]),
+            () => new ListBuilder<BuiltList<int>>())
+        ..addBuilderFactory(
+            const FullType(
+                BuiltMap, const [const FullType(int), const FullType(String)]),
+            () => new MapBuilder<int, String>())
+        ..addBuilderFactory(
+            const FullType(BuiltMap,
+                const [const FullType(String), const FullType(String)]),
+            () => new MapBuilder<String, String>())
+        ..addBuilderFactory(
+            const FullType(BuiltMap, const [
+              const FullType(BuiltMap,
+                  const [const FullType(int), const FullType(String)]),
+              const FullType(
+                  BuiltMap, const [const FullType(int), const FullType(String)])
+            ]),
+            () =>
+                new MapBuilder<BuiltMap<int, String>, BuiltMap<int, String>>()))
+      .build();
+
+  group('Serializers with StandardJsonPlugin', () {
+    test('throws on serialize if specifiedType is missing', () {
+      expect(() => serializers.serialize(1), throws);
+    });
+
+    test('throws on deserialize if specifiedType is missing', () {
+      expect(() => serializers.deserialize(1), throws);
+    });
+
+    test('throws on serialize of list multimaps', () {
+      final data = new BuiltListMultimap<int, String>({
+        1: ['one'],
+        2: ['two'],
+        3: ['three']
+      });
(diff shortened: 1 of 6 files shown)


brianegan:
Nice one!

built_value/test/standard_json_plugin_test.dart
```diff
@@ -0,0 +1,190 @@
...
+              const FullType(
+                  BuiltMap, const [const FullType(int), const FullType(String)])
+            ]),
+            () =>
+                new MapBuilder<BuiltMap<int, String>, BuiltMap<int, String>>()))
+      .build();
+
+  group('Serializers with StandardJsonPlugin', () {
+    test('throws on serialize if specifiedType is missing', () {
+      expect(() => serializers.serialize(1), throws);
```
brianegan: One Q: Why even allow this error to happen? Is `specifiedType` required in all cases, or are there times where it isn't required? 

As a consumer, if it's always required, I feel like it would make more sense to avoid runtime errors and simply make `specifiedType` a second, non-optional `specifiedType` param. If you still want a check, you could throw in an `assert(specifiedType != null)`.

Or did I totally miss the point of this one?

davidmorgan: The default serialization can handle serializing without specifiedType; this means it has to write the type name on the wire so that it can be deserialized correctly.

But the serialization added with this pull request always requires a specifiedType because it never writes type names on the wire. This means that, as you point out, the interface for 'serializers' is not ideal for this case. It would make sense to wrap it.

Someone suggested adding a way to do map-based serialization that also writes types on the wire; I expect to add this as another plugin you can choose. Then you won't always need specifiedType, but you'll need something that can understand the type names.

manuelF:
:lgtm:

---

Reviewed 10 of 10 files at r1.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[built_value/lib/standard_json_plugin.dart, line 68 at r1](https://reviewable.io:443/reviews/google/built_value.dart/89#-KabR2WVzj8Hjs36lybR:-KabR2WVzj8Hjs36lybS:bwtm1n8) ([raw file](https://github.com/google/built_value.dart/blob/ba6aefc03078118d8cc363dbc0cfbf3ce5e2a112/built_value/lib/standard_json_plugin.dart#L68)):*
> ```Dart
>     map.forEach((key, value) {
>       result[i++] = alreadyStringKeys ? key : _fromStringKey(key);
>       result[i++] = value;
> ```

nit: 
(key,value) {
result[i] = alreadyStringKeys...
result[i+1] = value
i += 2;
}
Looks cleaner to me. This way it is clearer that you are partitioning the things into (k, v), (k, v), (k, v).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/built_value.dart/89#-:-KabTRD1OOQXpqYNgIOm:bnfp4nl)*

davidmorgan:
Review status: 3 of 6 files reviewed at latest revision, 2 unresolved discussions.

---

*[built_value/lib/standard_json_plugin.dart, line 68 at r1](https://reviewable.io:443/reviews/google/built_value.dart/89#-KabR2WVzj8Hjs36lybR:-KabviQbMa6wYa7QQQq_:b-896fix) ([raw file](https://github.com/google/built_value.dart/blob/ba6aefc03078118d8cc363dbc0cfbf3ce5e2a112/built_value/lib/standard_json_plugin.dart#L68)):*
<details><summary><i>Previously, manuelF (Manuel Ferreria) wrote…</i></summary><blockquote>

nit: 
(key,value) {
result[i] = alreadyStringKeys...
result[i+1] = value
i += 2;
}
Looks cleaner to me. This way it is clearer that you are partitioning the things into (k, v), (k, v), (k, v).
</blockquote></details>

Done.

---

*[built_value/test/standard_json_plugin_test.dart, line 42 at r1](https://reviewable.io:443/reviews/google/built_value.dart/89#-KZRze136dSQdT_TmGpM-r1-42:-KabvmOxkfBGqbuHmRrq:b-896fix) ([raw file](https://github.com/google/built_value.dart/blob/ba6aefc03078118d8cc363dbc0cfbf3ce5e2a112/built_value/test/standard_json_plugin_test.dart#L42)):*
<details><summary><i>Previously, davidmorgan (David Morgan) wrote…</i></summary><blockquote>

The default serialization can handle serializing without specifiedType; this means it has to write the type name on the wire so that it can be deserialized correctly.

But the serialization added with this pull request always requires a specifiedType because it never writes type names on the wire. This means that, as you point out, the interface for 'serializers' is not ideal for this case. It would make sense to wrap it.

Someone suggested adding a way to do map-based serialization that also writes types on the wire; I expect to add this as another plugin you can choose. Then you won't always need specifiedType, but you'll need something that can understand the type names.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/built_value.dart/89)*

PR #113, opened 2017-01-30
update to build_runner 0.3.0

jakemac53:
This change is [Reviewable](https://reviewable.io/reviews/google/built_value.dart/113)


jakemac53:
If you could publish a new version after this that would be great, as the newer versions of build resolve a pretty major issue for at least mac, and possibly windows (https://github.com/dart-lang/build/issues/175).

PR #139, opened 2017-03-03
Allow users to define any suitable factory.

### LINKED ISSUE #108: Allow factory constructor to be named
The buit_value generator requires the class to have a factory constructors declared like this:

```dart
factory MyValue([updates(MyValueBuilder b)]) = _$MyValue;
```

I would like to allow the factory to be a named constructor.
It would allow us to use the default name to write a more user friendly constructor:

```dart
  // Required for BuiltValue
  factory MyValue._fromBuilder([updates(MyValueBuilder b)]) = _$MyValue;
  MyValue._();

  // Custom user-friendly constructor
  factory MyValue(num value1, {@required num value2}) {
    return new MyValue._fromBuilder((b) => b
      ..value1 = value1
      ..value2 = value2);
  }
```

davidmorgan:
Fixes #108 

This is actually slightly more than was asked for in that PR. It turns out that nothing depends on that factory. So, this PR makes any factory allowed. To help point users in the right direction, though, they're required to add a factory that references the generated type. (Otherwise their class will be unusable).

---
This change is [Reviewable](https://reviewable.io/reviews/google/built_value.dart/139)


manuelF:
:lgtm:

---

Reviewed 11 of 11 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/built_value.dart/139#-:-KeJgXQWg1VzIfS1wQ_Q:bnfp4nl)*

PR #143, opened 2017-03-07
Add serializer for "DateTime" fields.

### LINKED ISSUE #111: Serialisation of Dart DateTime
I am sure this issue has been covered some where was just not able to find it in the closed issues.

In an attempt to keep things simple I tend to use ```new DateTime.now()``` relative frequently directly in the builders, it works find because I am often used DateTime to calculate difference in relatively short time period.

To handle serialisation of this I could go the whole ```specifiedType = const FullType```  route and build a custom serialised for DateTime. But just feeling a bit over an over kill. Alternative of defining DateTime as an int and doing inline convert when I want to use ```DateTime``` or ```Duration``` is also messy. (Serialisation of DateTime is always a pain, I am happy with int or string in ISO format going over the wired).

Any ideas? I think if I my model centralised (one namespace/library) the custom serializer would be  ok, after all it is just one type. But that is not my use case. I have different model parts located in different isolates and I have DateTime all over the place.

Your guidance is appreciated... (not a blocking issues, just getting BuiltValue and Collection and finding it so good, I am probably over driven to find the ideal solution.)

davidmorgan:
Fixes #111 

This change is [Reviewable](https://reviewable.io/reviews/google/built_value.dart/143)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.5.6
+
+- Add serializer for "DateTime" fields.
+
 ## 0.5.5
 
 - Add serializer for "num" fields.
--- a/benchmark/pubspec.yaml
+++ b/benchmark/pubspec.yaml
@@ -12,12 +12,16 @@ environment:
 dependencies:
   browser: any
   built_collection: ^1.0.0
-  built_value: ^0.5.5
+# built_value: ^0.5.5
+  built_value:
+    path: ../built_value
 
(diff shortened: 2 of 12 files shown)


manuelF:
:lgtm:

---

Review status: 0 of 12 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/built_value.dart/143#-:-KedbXLN1i9sjIc5ne0j:bnfp4nl)*

PR #164, opened 2017-04-12
Add "built_value_test" library with matcher.

### LINKED ISSUE #156: Add Matcher support
For tests, it should be possible to compare field by field and output a specific error message on mismatch.

davidmorgan:
Fixes #156

---
This change is [Reviewable](https://reviewable.io/reviews/google/built_value.dart/164)

### DIFF
--- a/built_value_test/LICENSE
+++ b/built_value_test/LICENSE
@@ -1,0 +1,28 @@
+Copyright 2017, Google Inc. All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are
+met:
+
+   * Redistributions of source code must retain the above copyright
+notice, this list of conditions and the following disclaimer.
+   * Redistributions in binary form must reproduce the above
+copyright notice, this list of conditions and the following disclaimer
+in the documentation and/or other materials provided with the
+distribution.
+
+   * Neither the name of Google Inc. nor the names of its
+contributors may be used to endorse or promote products derived from
+this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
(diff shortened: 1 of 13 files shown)


built_value_test/lib/matcher.dart
```diff
@@ -0,0 +1,86 @@
...
+  @override
+  bool matches(dynamic item, Map matchState) {
+    if (_expected.runtimeType != item.runtimeType) return false;
+
+    return _delegate.matches(_toMap(item), matchState);
+  }
+}
+
+/// Converts a Built to a map.
+Map _toMap(Object built) {
```
manuelF: Isn't this just a Map<String, String> ?

built_value_test/lib/matcher.dart
```diff
@@ -0,0 +1,86 @@
...
+
+  // Reset to what it was on method entry.
+  newBuiltValueToStringHelper = previousNewBuiltValueToStringHelper;
+
+  return capturingToStringHelper.map;
+}
+
+/// Captures values in a Map instead of converting to a String.
+class _CapturingToStringHelper implements BuiltValueToStringHelper {
+  Map map = <String, Object>{};
```
manuelF: Map on LHS overrides the type from <string, Object>. Why not final?

built_value_test/LICENSE
```diff
@@ -0,0 +1,28 @@
+Copyright 2015, Google Inc. All rights reserved.
```
manuelF: Is this the right date?

built_value_test/lib/matcher.dart
davidmorgan: It's <String, Object>. Done.

(If it was <String, String> we'd end up checking for string equality instead of actual equality.)

built_value_test/lib/matcher.dart
davidmorgan: Done.

built_value_test/LICENSE
davidmorgan: Done.

PR #168, opened 2017-05-29
Significantly improve build performance by using @memoized instead of precomputed fields.

davidmorgan:
Straightforward refactor with no change to functionality or tests: instead of precomputing everything in e.g. ValueSourceClass constructor, store analyzer elements and do computation on them only in @memoized getters.

Significantly faster at build time when there is irrelevant code visible to the analyzer.

---
This change is [Reviewable](https://reviewable.io/reviews/google/built_value.dart/168)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 1.1.2
+
+- Significantly improve build performance by using @memoized instead of precomputed fields.
+
 ## 1.1.1
 
 - Update analyzer and build dependencies.
--- a/benchmark/pubspec.yaml
+++ b/benchmark/pubspec.yaml
@@ -12,12 +12,16 @@ environment:
 dependencies:
   browser: any
   built_collection: ^1.0.0
-  built_value: ^1.1.1
+# built_value: ^1.1.1
+  built_value:
+    path: ../built_value
 
 dev_dependencies:
   build: ^0.9.0
   build_runner: ^0.3.2
-  built_value_generator: ^1.1.1
+# built_value_generator: ^1.1.1
+  built_value_generator:
+    path: ../built_value_generator
   source_gen: '>=0.5.0+03 <0.6.0'
   quiver: '>=0.21.0 <0.26.0'
   test: any
--- a/built_value_generator/lib/built_value_generator.dart
+++ b/built_value_generator/lib/built_value_generator.dart
@@ -18,13 +18,11 @@ class BuiltValueGenerator extends Generator {
   @override
   Future<String> generate(Element element, BuildStep buildStep) async {
     if (element is ClassElement && ValueSourceClass.needsBuiltValue(element)) {
-      return new ValueSourceClass.fromClassElement(element).generateCode();
+      return new ValueSourceClass(element).generateCode();
     } else if (element is LibraryElement) {
-      final enumCode =
-          new EnumSourceLibrary.fromLibraryElement(element).generateCode();
+      final enumCode = new EnumSourceLibrary(element).generateCode();
 
-      final serializerSourceLibrary =
-          SerializerSourceLibrary.fromLibraryElement(element);
+      final serializerSourceLibrary = new SerializerSourceLibrary(element);
       if (serializerSourceLibrary.needsBuiltJson ||
           serializerSourceLibrary.hasSerializers) {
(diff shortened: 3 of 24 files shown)


manuelF:
Review status: 0 of 24 files reviewed at latest revision, 2 unresolved discussions.

---

*[built_value_generator/lib/src/serializer_source_class.dart, line 39 at r1](https://reviewable.io:443/reviews/google/built_value.dart/168#-KlIlufHYH4A7ryx-PEP:-KlIlufIVVhmxAdvFTHl:b-kz7qki) ([raw file](https://github.com/google/built_value.dart/blob/1c5e727388a3139b0176233d928c010d6fb176c8/built_value_generator/lib/src/serializer_source_class.dart#L39)):*
>   @memoized
>   BuiltList<String> get genericParameters => new BuiltList<String>(
>       element.typeParameters.map((e) => e.computeNode().toString()));

How do these work with memoized? The getter returns always a new instance of a BuiltList. With memoized will it return the same one? What about when/if element changes? (I don't thing it can change, but then why this isn't final? So we defer computation to actually when it accessed?)

---

*[built_value_generator/lib/src/serializer_source_field.dart, line 51 at r1](https://reviewable.io:443/reviews/google/built_value.dart/168#-KlIn3Ems6enJ3FdG0z4:-KlIn3Ems6enJ3FdG0z5:b-i1uut0) ([raw file](https://github.com/google/built_value.dart/blob/1c5e727388a3139b0176233d928c010d6fb176c8/built_value_generator/lib/src/serializer_source_field.dart#L51)):*
> 'nullable'
@nullable annotation defined in built_value is already a string. Can you just do:
metadata.constantValue == nullable

?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/built_value.dart/168)*

davidmorgan:
Review status: 0 of 24 files reviewed at latest revision, 2 unresolved discussions.

---

*[built_value_generator/lib/src/serializer_source_class.dart, line 39 at r1](https://reviewable.io:443/reviews/google/built_value.dart/168#-KlIlufHYH4A7ryx-PEP:-KlIpYbRov9o33_7w8Yr:b-ttnd9w) ([raw file](https://github.com/google/built_value.dart/blob/1c5e727388a3139b0176233d928c010d6fb176c8/built_value_generator/lib/src/serializer_source_class.dart#L39)):*
<details><summary><i>Previously, manuelF (Manuel Ferreria) wrote…</i></summary><blockquote>

>   @memoized
>   BuiltList<String> get genericParameters => new BuiltList<String>(
>       element.typeParameters.map((e) => e.computeNode().toString()));

How do these work with memoized? The getter returns always a new instance of a BuiltList. With memoized will it return the same one? What about when/if element changes? (I don't thing it can change, but then why this isn't final? So we defer computation to actually when it accessed?)
</blockquote></details>

With memoized it's only calculated once, on first access. Then it's stored and returned for all future calls.

So, yes, it always returns the same instance. The underlying element can never change, so that's exactly what we want.

Right, it's memoized rather than final to defer computation. In particular, if there's no need to compute anything -- no work is done.

---

*[built_value_generator/lib/src/serializer_source_field.dart, line 51 at r1](https://reviewable.io:443/reviews/google/built_value.dart/168#-KlIn3Ems6enJ3FdG0z4:-KlIq4TBu9MbOWWjlYNO:bg9dl6j) ([raw file](https://github.com/google/built_value.dart/blob/1c5e727388a3139b0176233d928c010d6fb176c8/built_value_generator/lib/src/serializer_source_field.dart#L51)):*
<details><summary><i>Previously, manuelF (Manuel Ferreria) wrote…</i></summary><blockquote>

> 'nullable'
@nullable annotation defined in built_value is already a string. Can you just do:
metadata.constantValue == nullable

?
</blockquote></details>

The metadata 'constantValue' doesn't return the actual value; it returns something that could be any type of value and has accessor methods like 'toStringValue' for if you think it's supposed to be a string.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/built_value.dart/168)*

manuelF:
:lgtm:

---

Review status: 0 of 24 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/built_value.dart/168#-:-KlIqQixDreEhyluVgaF:bnfp4nl)*

Issue #178, opened 2017-06-22
built_value depends on dart 1.21+ for Generic Method syntax


jefftougas:
First off, thanks for this awesome library. I'm really digging it.

It looks like built_value.dart is incompatible with dart older than 1.21 due to the use of generics in a couple of convenience methods (deserializeWith and serializeWith in serializers.dart and built_json_serializers.dart). Does that sound right to you or am I missing something?

If my understanding is correct, then I will submit a PR to correct built_value's pubspec sdk requirement to reflect this (right now it calls for >1.8.0), and also I'll create a branch in my fork for legacy dart support that simply removes those convenience methods whilst leaving the sdk requirement as is. If you'd be willing to incorporate said legacy support branch into your repo and perhaps publish a patch on pub to one of the older major versions (or similar), that'd be really helpful as it'd allow me to go on referencing the main pub repo from my pubspec.yaml instead of pointing to my fork / branch, but obviously no worries if you don't think that makes sense for whatever reason.

davidmorgan:
Thanks for catching. Yes, the pubspec should be updated -- please do send a PR.

If I understand correctly what you'd like is for a version to be published to pub with the generics moved back to comments? Makes sense. Do you need it to be on github or is just the pub package enough? I'll take a look on Monday.

jefftougas:
> Yes, the pubspec should be updated -- please do send a PR.

Thanks for confirming - will send in a PR shortly.

> If I understand correctly what you'd like is for a version to be published to pub with the generics moved back to comments? Makes sense. Do you need it to be on github or is just the pub package enough?

I had in mind commenting out (or removing) the 4 methods that use generics, but perhaps you're referring to [this functionality](https://github.com/dart-archive/dev_compiler/blob/master/doc/GENERIC_METHODS.md#using-generic-method-parameters)? In either case, yes, a version being published to pub that *somehow* is compatible with older dart is my ideal outcome here, and the pub package would be sufficient - no need to publish to github in that scenario.

jefftougas:
PR opened. Also FYI, I pushed the commented generic syntax to my fork with 64ee528, which I'll be using in my dart 1.15 project until you get a chance to publish a special version with a similar change to pub. Thanks again!

davidmorgan:
This should now be done -- release v1.1.3 with the old syntax and v1.1.4 with the new syntax and the SDK constraint.

Please let me know if you have any problems.

Thanks!

jefftougas:
[...]

PR #179, opened 2017-06-23
Update minimum dart sdk to reflect usage of method generics

jefftougas:
Background: https://github.com/google/built_value.dart/issues/178

---
This change is [Reviewable](https://reviewable.io/reviews/google/built_value.dart/179)

### DIFF
--- a/built_value/pubspec.yaml
+++ b/built_value/pubspec.yaml
@@ -8,7 +8,7 @@ authors:
 homepage: https://github.com/google/built_value.dart
 
 environment:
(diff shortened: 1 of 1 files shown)


jefftougas:
I signed it!

davidmorgan:
Thanks -- I realized the way to do this correctly was to switch to the old syntax, release, then bump the version and add the SDK syntax, and publish again. So I went ahead and did those; no need for this PR any more.

Thanks for pointing out the issue!

PR #186, opened 2017-07-04
Polymorphism support: pick up fields from interfaces, add BuiltValue annotation, add non-instantiable value types.

### LINKED ISSUE #22: Pick up properties from implemented interfaces
Unless I'm missing something you can't currently have a hierarchical model, where you inherit properties from parent models and similarly builders inherit from other builders.

Is this planned?

davidmorgan:
Fixes #22

---
This change is [Reviewable](https://reviewable.io/reviews/google/built_value.dart/186)


end_to_end_test/test/polymorphism_test.dart
```diff
@@ -0,0 +1,75 @@
...
+        new Cat((b) => b
+          ..legs = 4
+          ..tail = true),
+        new Fish((b) => b
+          ..legs = 0
+          ..fins = 2),
+      ];
+
+      final modifiedAnimals =
+          animals.map((animal) => animal.rebuild((b) => b.legs++)).toList();
```
filiph: This is _so_ good. Any way to show it in examples and/or docs?

I know that, rationally, there's no reason to think this wouldn't work. But experience has trained people like me to assume otherwise.

example/lib/polymorphism.dart
```diff
@@ -0,0 +1,72 @@
...
+
+  factory Fish([updates(FishBuilder b)]) = _$Fish;
+  Fish._();
+}
+
+/// As with any Built Value, behaviour can be added via mixins.
+abstract class Walker {
+  int get legs;
+
+  bool get canWalk => legs > 0;
```
filiph: [flame war:] One-legged walker doesn't walk, it _hops_. Only 2+ legs can walk.

CHANGELOG.md
```diff
@@ -1,8 +1,12 @@
 # Changelog
 
-## 1.1.5
+## 1.2.0
 
 - Fix depending on a fully generated builder from a manually maintained builder.
+- Pick up fields on implemented interfaces, so you don't have to @override them.
+- Add BuiltValue annotation for specifying settings.
```
filiph: Reading just this changelog without the example isn't very helpful. Yet, users want to learn about new things from the CHANGELOG. 

#### Option 1: build changelogs that include example usage, [like the Dart SDK does](https://github.com/dart-lang/sdk/blob/master/CHANGELOG.md#1240---12-06-2017), for example.

* Pros: Right there in the changelog. Doesn't need to update because in the changelog, example clearly attaches to a point in time.
* Cons: More work. Can make CHANGELOGs a bit unwieldy.

#### Option 2: add a canonical example or two to README.md and link to the section from here

* Pros: Example usage is easy to find and read. It's just one example so clear to know what to look at (contrast with many files in the `example/` dir). Way to advertise features to new users.
* Cons: Needs updating. Breaking changes to the example won't get auto-detected by analyzer.

#### Option 3: link from changelog to example files

* Pros: Least amount of immediate work. 
* Cons: Links can get stale or broken.

example/lib/polymorphism.dart
RollyPeres: @filiph @davidmorgan  How is it possible to force a class using a mixin to override a getter provided by this mixin? Something similar to how `AutomaticKeepAliveClientMixin` forces you to override `wantKeepAlive`.

filiph: Not sure if I remember the question correctly, but here goes:

```dart
mixin AutomaticKeepAliveClientMixin {
  void foo() {
    // actually do something useful here
  }

  bool get wantKeepAlive;
}
```

Now `wantKeepAlive` is abstract and when you mix in `AutomaticKeepAliveClientMixin` into a non-abstract class, you should be forced to implement `wantKeepAlive`.

RollyPeres: @filiph I was too tired to notice the class I was mix in my mixin was abstract, so of course it's not going to yell for an override. Thought there was more to it, but it's pretty straight forward. Thanks!

PR #193, opened 2017-07-24
Upgrade to source_gen 0.7.0

natebosch:
Use a git dependency for now since it is unpublished.

Since `Generator` runs per library, emulate the old behavior by
iterating over elements manually, writing error messages to the
source, and logging exceptions at `severe`.

---
This change is [Reviewable](https://reviewable.io/reviews/google/built_value.dart/193)

### DIFF
--- a/benchmark/pubspec.yaml
+++ b/benchmark/pubspec.yaml
@@ -12,12 +12,16 @@ environment:
 dependencies:
   browser: any
   built_collection: ^1.0.0
-  built_value: ^1.2.1
+# built_value: ^1.2.1
+  built_value:
+    path: ../built_value
 
 dev_dependencies:
   build: ^0.9.0
   build_runner: ^0.3.2
-  built_value_generator: ^1.2.1
+# built_value_generator: ^1.2.1
+  built_value_generator:
+    path: ../built_value_generator
   source_gen: ^0.7.0
   quiver: '>=0.21.0 <0.26.0'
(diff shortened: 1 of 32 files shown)


natebosch:
Sending early for feedback - will need to wait until `source_gen` is published.

In the long term we probably want to refactor the tests so the don't rely on the per-element behavior. `source_gen` used to be very lenient about letting builds through and depending on the build system to fail on `log.severe`, but in the long term we want any failure to be surfaced directly rather than written into the output as a comment.

dave26199:
Thanks -- will take a look on Monday when I'm back from vacation.

natebosch:
As a heads up I'll likely need to patch this internally before then - we can resolve the delta next week.

dave26199:
Sounds good -- whatever gets you unblocked. Thanks.

davidmorgan:
Looks good, thanks.

To get the presubmit to pass, please run:

tool/use_local_deps

Which makes all the internal dependencies (built_value_generator->built_value, etc) use paths. I'll handle publishing a new version to pub -- at which point local paths are no longer needed.

Thanks!

natebosch:
Travis status seems to be stuck... but the checks passed if you follow the link.

Let me know if there is anything else I can help with.

davidmorgan:
Looks good -- thanks for working through this in my absence, much appreciated.

PR #202, opened 2017-08-02
Add nestedBuilders setting.

### LINKED ISSUE #199: Add setting for turning off nested builders
By default if the builder is fully generated we created nested builders.

Add a setting to instead default to no nested builders.

davidmorgan:
Fixes #199

---
This change is [Reviewable](https://reviewable.io/reviews/google/built_value.dart/202)


built_value_generator/lib/built_value_generator.dart
```diff
@@ -27,9 +27,12 @@ class BuiltValueGenerator extends Generator {
           serializerSourceLibrary.hasSerializers) {
         result.writeln(serializerSourceLibrary.generateCode());
       }
-    } catch (e, st) {
+    } on InvalidGenerationSourceError catch (e, st) {
```
manuelF: Aren't both handle cases exactly identical?

davidmorgan: The difference is whether we call "e.message"; only InvalidGenerationSourceError has a "message".

PR #215, opened 2017-09-06
Find serializable classes via an annotation instead of searching all libraries.

davidmorgan:
This change is [Reviewable](https://reviewable.io/reviews/google/built_value.dart/215)


built_value_generator/lib/src/serializer_source_library.dart
```diff
@@ -22,20 +22,33 @@ abstract class SerializerSourceLibrary
...
+  bool get hasSerializers => serializersForAnnotation != null;
+
+  /// Returns the @serializersFor annotation from an accessor called
+  /// "serializers", or null if none is found.
+  @memoized
+  ElementAnnotation get serializersForAnnotation {
+    final accessors = element.definingCompilationUnit.accessors
+        .where((element) =>
+            element.isGetter && element.displayName == 'serializers')
+        .toList();
```
manuelF: Unnecessary .toList

davidmorgan: Without the "toList" we iterate all the accessors twice -- once for "isEmpty" and once for "single". So I think it's nicer to call "toList" here.

manuelF: Technically no, iterable isEmpty is implemented by checking if you can move an iterator, so O(1) as well.
https://api.dartlang.org/stable/1.24.2/dart-core/Iterable/isEmpty.html

But it is just a minor nit.

davidmorgan: There "where" means that to answer "isEmpty" it has to find the first match -- or iterate over all elements if there aren't any.

PR #268, opened 2017-11-24
Add a recipe for boilerplate generation in IntelliJ

ddrone:
I've added a live template for generating bits of boilerplate required to use built_value and decided to share it in the docs. Not sure whether README.md is the best place for that but there doesn't seem to be other places to put docs to?

---
This change is [Reviewable](https://reviewable.io/reviews/google/built_value.dart/268)


davidmorgan:
Nice. Thanks!

PR #270, opened 2017-11-26
Add ignored annotation for fields that should not be compared

ralkan:
I've had an issue with using flutter_redux and when creating a ViewModel for the store to pass on to the widget, I wasn't able to compare the old ViewModel instance with the new one every time a change happened. Since I had to pass a `String code` and a `Function onChange` (for dispatching actions) field in the ViewModel.

And when the field `code` is being compared it will return `true` however because the `onChange` field is a closure function, it always compares as `false` so the widget gets updated even when there is no direct change for this widget in the state.

```
(store) => new RoomTypeDropdownViewModel((b) => b
      ..code = store.state.roomTypeCode
      ..onChanged = (v) => store.dispatch({
        'type': SearchRoomTypeActions.Select,
        'room_type': v
      }));
```

I think we need a way to make `bool operator ==(other)` ignore specific fields.

With this PR we will be able to ignore certain fields in `bool operator ==(other)`.

---
This change is [Reviewable](https://reviewable.io/reviews/google/built_value.dart/270)

### DIFF
--- a/built_value/lib/built_value.dart
+++ b/built_value/lib/built_value.dart
@@ -88,6 +88,11 @@ const String nullable = 'nullable';
 
 /// Memoized annotation for Built Value getters and methods.
 ///
+/// Fields marked with this annotation are not compared in the == operator.
+const String ignored = 'ignored';
+
+/// Memoized annotation for Built Value getters and methods.
+///
 /// Getters marked with this annotation are memoized: the result is calculated
 /// once on first access and stored in the instance.
 const String memoized = 'memoized';
--- a/built_value/lib/built_value.dart
+++ b/built_value/lib/built_value.dart
@@ -86,10 +86,13 @@ class BuiltValue {
 /// Fields marked with this annotation are allowed to be null.
 const String nullable = 'nullable';
 
-/// BuiltValueField annotation for Built Value fields.
-///
-/// Fields marked with this annotation are not compared in the == operator.
+/// Optionally, annotate a Built Value field with this to specify settings.
+/// This is only needed for advanced use.
 class BuiltValueField {
+  /// Whether the field is compared and hashed. Defaults to `true`.
+  ///
+  /// Set to `false` to ignore the field when calculating `hashCode` and when
+  /// comparing with `operator==`.
   final bool compare;
 
   const BuiltValueField({this.compare: true});
--- a/built_value/lib/built_value.dart
+++ b/built_value/lib/built_value.dart
@@ -86,10 +86,14 @@ class BuiltValue {
 /// Fields marked with this annotation are allowed to be null.
 const String nullable = 'nullable';
 
-/// Memoized annotation for Built Value getters and methods.
+/// BuiltValueField annotation for Built Value fields.
 ///
 /// Fields marked with this annotation are not compared in the == operator.
-const String ignored = 'ignored';
+class BuiltValueField {
+  final bool compare;
+
+  const BuiltValueField({this.compare: true});
+}
 
 /// Memoized annotation for Built Value getters and methods.
 ///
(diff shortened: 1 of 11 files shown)


built_value/lib/built_value.dart
```diff
@@ -86,6 +86,11 @@ class BuiltValue {
 /// Fields marked with this annotation are allowed to be null.
 const String nullable = 'nullable';
 
+/// Memoized annotation for Built Value getters and methods.
+///
+/// Fields marked with this annotation are not compared in the == operator.
+const String ignored = 'ignored';
```
zoechi: I think the name should be more specific about what the field should be ignored for.
I mentioned ignore for serialization a while ago for example

ralkan: You're absolutely right! But, unfortunately I couldn't find any other word to describe the intent. Do you maybe have any suggestions?

ralkan:
Comment for the CLA bot to run again.

davidmorgan:
(Will review tomorrow). You can add yourself to AUTHORS if you like :)

ralkan:
@davidmorgan Great! Looking forward to your review.
I'd love to add myself to AUTHORS. Thanks! :)

built_value/lib/built_value.dart
zoechi: `ignoreInEquals`?

built_value/lib/built_value.dart
```diff
@@ -86,10 +86,14 @@ class BuiltValue {
 /// Fields marked with this annotation are allowed to be null.
 const String nullable = 'nullable';
 
-/// Memoized annotation for Built Value getters and methods.
+/// BuiltValueField annotation for Built Value fields.
```
davidmorgan: Let's mimic the style above.

```
/// Optionally, annotate a Built Value field with this to specify settings. This is only
/// needed for advanced use.
```

built_value/lib/built_value.dart
```diff
@@ -86,10 +86,14 @@ class BuiltValue {
...
 ///
 /// Fields marked with this annotation are not compared in the == operator.
```
davidmorgan: You don't need this line any more.

built_value/lib/built_value.dart
```diff
@@ -86,10 +86,14 @@ class BuiltValue {
...
-const String ignored = 'ignored';
+class BuiltValueField {
+  final bool compare;
```
davidmorgan: Please add comment to this field:

```
/// Whether the field is compared and hashed. Defaults to `true`.
///
/// Set to `false` to ignore the field when calculating `hashCode` and when
/// comparing with `operator==`.

File: built_value_generator/lib/src/value_source_field.dart [hunk 5]
```diff
@@ -59,8 +59,10 @@ abstract class ValueSourceField
...
 
   @memoized
-  bool get isIgnored => element.getter.metadata
-      .any((metadata) => metadataToStringValue(metadata) == 'ignored');
+  bool get isCompared => element.getter.metadata.every((metadata) =>
+      getMetadataField(metadata, 'compare') != null
+          ? getMetadataField(metadata, 'compare').toBoolValue()
+          : true);
 
   @memoized
```
davidmorgan: This is currently checking the `compare` field on every annotation; I think what we want is to pull out the BuiltValueField value once, and check that here.

You can follow the pattern for the BuiltValue annotation:

https://github.com/google/built_value.dart/blob/master/built_value_generator/lib/src/value_source_class.dart#L37

ralkan: You're right. That looks much better. I'll push the updated code in a bit. 👍

built_value/lib/built_value.dart
ralkan: Thanks for your suggestion @zoechi.
We went with `BuiltValueField(compare: false)` as suggested by @davidmorgan.

ralkan:
@davidmorgan Done! 👍  I pushed a new commit with the latest changes. :)

davidmorgan:
Looks good, thanks!

PR #286, opened 2017-12-18
Fail with error on public fields.

### LINKED ISSUE #277: Handle private fields
Should private fields be allowed? If not, they should be explicitly disallowed.

davidmorgan:
Fixes #277

---
This change is [Reviewable](https://reviewable.io/reviews/google/built_value.dart/286)


manuelF:
What about private methods? Are they disallowed as well?

davidmorgan:
No, they're fine.

Private fields already didn't work, this just gives them a helpful error message instead of a compiler error :)

PR #291, opened 2017-12-18
Support skipping fields when serializing.

### LINKED ISSUE #207: Add an annotation to indicate a field should be skipped for serialization
Perhaps a custom serialization plugin would be able to do that as well, but without annotation support, it is cumbersome.

davidmorgan:
Fixes #207

---
This change is [Reviewable](https://reviewable.io/reviews/google/built_value.dart/291)


built_value/lib/built_value.dart
```diff
@@ -95,7 +95,13 @@ class BuiltValueField {
   /// comparing with `operator==`.
   final bool compare;
 
-  const BuiltValueField({this.compare: true});
+  /// Whether the field is serialized. Defaults to `true`.
+  ///
+  /// If a field is not serialized, it must either be `@nullable` or specify a
+  /// default for deserialization to succeed.
```
zoechi: specify a default value how?

davidmorgan: Currently I believe this has to be by hand-coding the builder.

zoechi: ok, thanks for the feedback

naiveai:
It's still not really specified anywhere where to set default values such that it will succeed when a deserializer attempts to come create a class - is there some piece of documentation I'm missing?

MichaelM97:
> It's still not really specified anywhere where to set default values such that it will succeed when a deserializer attempts to come create a class - is there some piece of documentation I'm missing?

Just annotate the field with `@BuiltValueField(serialize: false)` :)

PR #295, opened 2017-12-19
Support overriding wireName for enums.

### LINKED ISSUE #144: Add serializer plugin that does renames
That is one of requirements,  that probably appears only when one works with external API.
In my case models have many String fields that can have values only from some predefined list:
For example, the field chartType can have values

-   table
-   barchart
-   linechart
-  pivot-table

and so on. Ideally, I would like to use Enums for such fields, but some of the possible values (pivot-table in above example) are not valid dart identifiers. 

It would be great If we could change serialized value of Enum by some annotation. 

For example, given enum 

``` Dart

class SecondTestEnum extends EnumClass {
  static const SecondTestEnum yes = _$ys;
  static const SecondTestEnum no = _$n;
  static const SecondTestEnum definitely = _$definitely;
  @SerializedAs('no-thanks')
  static const SecondTestEnum noThanks = _$noThanks;
  const SecondTestEnum._(String name) : super(name);
  static Serializer<SecondTestEnum> get serializer => _$secondTestEnumSerializer;
  static BuiltSet<SecondTestEnum> get values => _$vls;
  static SecondTestEnum valueOf(String name) => _$vlOf(name);
}

```
Its implementation would be:

``` Dart
const SecondTestEnum _$ys = const SecondTestEnum._('yes');
const SecondTestEnum _$n = const SecondTestEnum._('no');
const SecondTestEnum _$definitely = const SecondTestEnum._('definitely');
const SecondTestEnum _$noThanks = const SecondTestEnum._('no-thanks');

SecondTestEnum _$vlOf(String name) {
  switch (name) {
    case 'yes':
      return _$ys;
    case 'no':
      return _$n;
    case 'definitely':
      return _$definitely;
    case 'no-thanks':
      return _$noThanks;
    default:
      throw new ArgumentError(name);
  }
}

...
```

davidmorgan:
Fixes #144

---
This change is [Reviewable](https://reviewable.io/reviews/google/built_value.dart/295)

### DIFF
--- a/built_value_generator/lib/src/serializer_source_class.dart
+++ b/built_value_generator/lib/src/serializer_source_class.dart
@@ -8,6 +8,8 @@ import 'package:analyzer/dart/element/element.dart';
 import 'package:analyzer/dart/element/type.dart';
 import 'package:built_collection/built_collection.dart';
 import 'package:built_value/built_value.dart';
+import 'package:built_value_generator/src/enum_source_class.dart';
+import 'package:built_value_generator/src/enum_source_field.dart';
 import 'package:built_value_generator/src/fields.dart' show collectFields;
 import 'package:built_value_generator/src/serializer_source_field.dart';
 import 'package:built_value_generator/src/value_source_class.dart';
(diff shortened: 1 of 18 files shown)


built_value_generator/lib/src/serializer_source_class.dart
```diff
@@ -29,13 +31,21 @@ abstract class SerializerSourceClass
...
+  // TODO(davidmorgan): share common code in a nicer way.
+  @memoized
+  BuiltValueEnum get enumClassSettings => new EnumSourceClass(element).settings;
 
   @memoized
   String get name => element.name;
 
   @memoized
-  String get wireName => settings.wireName ?? name;
+  String get wireName =>
```
manuelF: If you use a { if... } it would be slightly more readable.

end_to_end_test/lib/enums.dart
```diff
@@ -10,21 +10,7 @@ import 'package:built_value/serializer.dart';
 
 part 'enums.g.dart';
 
-/// Example of how to use [EnumClass].
```
manuelF: Why do you remove the doc?

built_value_generator/lib/src/serializer_source_class.dart
davidmorgan: Yup. Done!

end_to_end_test/lib/enums.dart
davidmorgan: There aren't supposed to be docs in the 'end_to_end_test' folder; the same file exists over in the 'example' folder with docs.

But, actually, I should add an example. Done.

PR #296, opened 2017-12-19
Support polymorphism in StandardJsonPlugin.

### LINKED ISSUE #276: StandardJsonPlugin serialization of list of polymorphic objects
I want to use StandardJsonPlugin to generate a map I can save to a document database on the server side. It needs to be structured as a map so I can create indexes on the data in the database. However the objects I want to serialize are polymorphic.

So given a class hierarchy of animals like in the [built_value polymorphism example][1]:

    Animal
        - Cat
        - Fish

I want to be able to (de)serialize a list of animals to JSON

    class Zoo implements Built<Zoo, ZooBuilder>  {
        static Serializer<Zoo> get serializer => _$zooSerializer;
        BuiltList<Animal> animals;
    }

I would like the type of animal to be encoded in each item. For example the json output could look like:

    {
    "animals": [
        {"type": "cat", "legs": 2},
        {"type": "fish", "fins": 3}
    ]}

For the moment I am working around this by creating an intermediate object like and removing the polymorphic-ness of the objects:

    class AnAnimal {
        @nullable
        Cat cat;
        @nullable
        Fish fish;
    }

  [1]: https://github.com/google/built_value.dart/blob/0f131d8cedb271405a52652f9b5eaa61c2c589aa/example/lib/polymorphism.dart

davidmorgan:
Fixes #276

---
This change is [Reviewable](https://reviewable.io/reviews/google/built_value.dart/296)

### DIFF
--- a/built_value/lib/standard_json_plugin.dart
+++ b/built_value/lib/standard_json_plugin.dart
@@ -5,18 +5,25 @@ import 'dart:convert' show JSON;
 
 /// Switches to "standard" JSON format.
 ///
-/// The default serialization format is more powerful, supporting polymorphism
-/// and more collection types. But, you may need to interact with other systems
-/// that use simple map-based JSON. If so, use [SerializersBuilder.addPlugin]
-/// to install this plugin.
+/// The default serialization format is more powerful, with better performance
+/// and support for more collection types. But, you may need to interact with
+/// other systems that use simple map-based JSON. If so, use
+/// [SerializersBuilder.addPlugin] to install this plugin.
 class StandardJsonPlugin implements SerializerPlugin {
   static final BuiltSet<Type> _unsupportedTypes =
       new BuiltSet<Type>([BuiltSet, BuiltListMultimap, BuiltSetMultimap]);
 
+  /// The field used to specify the value type if needed. Defaults to `$`.
+  final String discriminator;
+
+  // The key used when there is just a single value, for example if serializing
+  // an `int`.
+  final String valueKey;
+
+  StandardJsonPlugin({this.discriminator: r'$', this.valueKey: ''});
+
   @override
   Object beforeSerialize(Object object, FullType specifiedType) {
-    if (specifiedType.isUnspecified)
-      throw new ArgumentError('Standard JSON requires specifiedType.');
     if (_unsupportedTypes.contains(specifiedType.root)) {
       throw new ArgumentError(
           'Standard JSON cannot serialize type ${specifiedType.root}.');
@@ -26,18 +33,30 @@ class StandardJsonPlugin implements SerializerPlugin {
 
   @override
   Object afterSerialize(Object object, FullType specifiedType) {
-    return object is List &&
-            specifiedType.root != BuiltList &&
-            specifiedType.root != JsonObject
-        ? _toMap(object, _alreadyHasStringKeys(specifiedType))
-        : object;
+    if (object is List &&
+        specifiedType.root != BuiltList &&
+        specifiedType.root != JsonObject) {
+      if (specifiedType.isUnspecified) {
+        return _toMapSpecifyType(object);
+      } else {
+        return _toMap(object, _alreadyHasStringKeys(specifiedType));
+      }
+    } else {
+      return object;
+    }
   }
 
   @override
   Object beforeDeserialize(Object object, FullType specifiedType) {
-    return object is Map && specifiedType.root != JsonObject
-        ? _toList(object, _alreadyHasStringKeys(specifiedType))
-        : object;
+    if (object is Map && specifiedType.root != JsonObject) {
+      if (specifiedType.isUnspecified) {
+        return _toListExtractType(object);
+      } else {
+        return _toList(object, _alreadyHasStringKeys(specifiedType));
+      }
+    } else {
+      return object;
+    }
   }
 
   @override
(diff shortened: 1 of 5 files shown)


davidmorgan:
Missed a few cases, let me take a look.

davidmorgan:
Now fixed, PTAL.

The JSON for primitive collections when the type is unknown is quite ugly--but users are actually unlikely to use this in practice. If the type is known (because the collection is wrapped in a class) then the collection doesn't need to be on the wire.

It's nice to have StandardJsonPlugin support all the polymorphism cases than the default serialization does, though; it now should do.

Thanks!

built_value/lib/standard_json_plugin.dart
```diff
@@ -59,6 +78,44 @@ class StandardJsonPlugin implements SerializerPlugin {
     return result;
   }
 
+  Map _toMapSpecifyType(List list) {
```
manuelF: _toMapSpecificType ?

built_value/lib/standard_json_plugin.dart
```diff
@@ -59,6 +78,44 @@ class StandardJsonPlugin implements SerializerPlugin {
...
+    var type = list[0];
+
+    if (list.length == 2) {
+      // Just a type and a primitive value. Encode the value in the map.
+      return <Object, Object>{discriminator: type, valueKey: list[1]};
+    }
+
+    if (list[0] == 'list') {
```
manuelF: if (type == 'list')

built_value/lib/standard_json_plugin.dart
```diff
@@ -5,18 +5,25 @@ import 'dart:convert' show JSON;
...
+/// The default serialization format is more powerful, with better performance
+/// and support for more collection types. But, you may need to interact with
+/// other systems that use simple map-based JSON. If so, use
+/// [SerializersBuilder.addPlugin] to install this plugin.
 class StandardJsonPlugin implements SerializerPlugin {
   static final BuiltSet<Type> _unsupportedTypes =
       new BuiltSet<Type>([BuiltSet, BuiltListMultimap, BuiltSetMultimap]);
 
+  /// The field used to specify the value type if needed. Defaults to `$`.
+  final String discriminator;
```
manuelF: Differentiator ? Slightly less bias.

built_value/lib/standard_json_plugin.dart
```diff
@@ -59,6 +78,44 @@ class StandardJsonPlugin implements SerializerPlugin {
     return result;
   }
 
+  Map _toMapSpecifyType(List list) {
+    var type = list[0];
+
+    if (list.length == 2) {
```
manuelF: can list.length == 1?

built_value/lib/standard_json_plugin.dart
davidmorgan: :D

I agree it's a bit aggressive sounding, but this appears to be the standard term

https://www.google.ch/search?q=json+discriminator

built_value/lib/standard_json_plugin.dart
davidmorgan: _toMapWithDiscriminator is a better name. Also went through and added comments to most of the methods in this class, since it's pretty complicated :) thanks.

built_value/lib/standard_json_plugin.dart
davidmorgan: Nope; the list contains at least a type name and a value. Added a comment.

built_value/lib/standard_json_plugin.dart
davidmorgan: Done.

davidmorgan:
Thanks, PTAL :)

built_value/lib/standard_json_plugin.dart
```diff
@@ -26,58 +33,170 @@ class StandardJsonPlugin implements SerializerPlugin {
...
+    if (type == 'list') {
+      return [type]..addAll(map[valueKey] as Iterable);
+    }
+
+    if (map.containsKey(valueKey)) {
+      // Just a type and a primitive value. Retrieve the value in the map.
+      final result = new List(2);
+      result[0] = type;
+      result[1] = map[valueKey];
+      return result;
```
manuelF: return [type, map[valueKey] ] ; ?

example/lib/example.dart
```diff
@@ -75,8 +75,13 @@ void example() {
...
 /// The plugin changes the built_value serialization format to the Map-based
-/// format used by most JSON APIs. You must specify which type you are
-/// serializing/deserializing when using this plugin.
+/// format used by most JSON APIs.
+///
+/// Where needed the plugin specifies which type to serialize/deserialize using
+/// a `discriminator` field. By default this field is called `$`; you can pass
+/// a different name to the `StandardJsonPlugin` constructor. If the wire
+/// format should not contain the type name, you must instead explicitly
+/// specify which type to serialize/deserialize.
```
manuelF: I can't find a proper example for this, can you make one? Easier to read than text I imagine.

built_value/lib/standard_json_plugin.dart
davidmorgan: I believe that would use more memory as it would return a growable list; as I have it, we get a fixed list of size two.

example/lib/example.dart
davidmorgan: Very much so. Done :)

PR #305, opened 2017-12-22
Make presubmit fail if formatting is wrong.

davidmorgan:
This change is [Reviewable](https://reviewable.io/reviews/google/built_value.dart/305)


davidmorgan:
Not merging as there seem to be mismatches on dartfmt; so it's not a good presubmit.

Issue #320, opened 2018-01-15
Consistent generator error messages


davidmorgan:
Two things: 1) always mention the name of the class with the problem, 2) always quote values rather than leaving them bare.

davidmorgan:
Some error messages can be confusing without quotes--let's do this one...

davidmorgan:
Also 3) make sure we don't print `dynamic` for the builder type in the error message.

PR #321, opened 2018-01-15
Prohibit sdk collections

davidmorgan:
This change is [Reviewable](https://reviewable.io/reviews/google/built_value.dart/321)

### DIFF
--- a/built_value_generator/lib/src/value_source_class.dart
+++ b/built_value_generator/lib/src/value_source_class.dart
@@ -256,11 +256,6 @@ abstract class ValueSourceClass
           'Only "implements" and "extends Object with" are allowed.');
     }
 
-    if (extendsNonObject) {
-      result.add('Stop class extending other classes. '
-          'Only "implements" and "extends Object with" are allowed.');
-    }
-
     if (settings.instantiable) {
       final expectedConstructor = '$name._()';
       if (valueClassConstructors.length != 1 ||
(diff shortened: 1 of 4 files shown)


built_value_generator/lib/src/value_source_class.dart
```diff
@@ -209,6 +249,11 @@ abstract class ValueSourceClass
       }
     }
 
+    if (extendsNonObject) {
+      result.add('Stop class extending other classes. '
```
zoechi: Looks like there is a "using" missing after "Stop"

davidmorgan: "Class" is supposed to refer to the value class, as opposed to the builder. But I agree it's unclear. I filed #320 to improve this, I think it would be better to always include the class/builder name.

I'm probably going to drop this PR anyway, need to split it into smaller pieces and do some related fixes first.

Thanks :)

zoechi: Ok, got it now. Was quite early in the morning when I looked though

built_value_generator/lib/src/value_source_field.dart
```diff
@@ -159,6 +167,11 @@ abstract class ValueSourceField
       result.add('Make field $name public; remove the underscore.');
     }
 
+    if (_suggestedTypes.keys.contains(type)) {
+      result.add('Make field "$name" have type "${_suggestedTypes[type]}". '
+          'The current type, "$type", is not allowed.');
```
manuelF: how about: "The current mutable type..." ?
Might be a better explanation to switch

davidmorgan: Done, "is not allowed because it is mutable."

Thanks!

PR #332, opened 2018-01-23
Prohibit importing built_value using 'show' or 'as'.

davidmorgan:
This change is [Reviewable](https://reviewable.io/reviews/google/built_value.dart/332)

### DIFF
--- a/built_value_generator/lib/src/value_source_class.dart
+++ b/built_value_generator/lib/src/value_source_class.dart
@@ -15,6 +15,11 @@ import 'package:source_gen/source_gen.dart';
 
 part 'value_source_class.g.dart';
 
+const String _importWithSingleQuotes =
+    "import 'package:built_value/built_value.dart'";
+const String _importWithDoubleQuotes =
+    'import "package:built_value/built_value.dart"';
+
 abstract class ValueSourceClass
     implements Built<ValueSourceClass, ValueSourceClassBuilder> {
   ClassElement get element;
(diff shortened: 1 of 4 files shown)


built_value_generator/lib/src/value_source_class.dart
```diff
@@ -99,7 +108,25 @@ abstract class ValueSourceClass
...
+    return source.contains(expectedCode);
+  }
+
+  @memoized
+  bool get hasBuiltValueImportWithShow {
+    // It would be more accurate to check using the AST, but this is
+    // potentially expensive. We already have the source for the "part of"
+    // check, use that.
+    return source.contains('$_importWithSingleQuotes show') ||
+        source.contains('$_importWithSingleQuotes show');
```
manuelF: _importWithDoubleQuotes

built_value_generator/lib/src/value_source_class.dart
```diff
@@ -203,6 +230,16 @@ abstract class ValueSourceClass
       result.add('Make class abstract.');
     }
 
+    if (hasBuiltValueImportWithShow) {
+      result.add('Stop using "show" when importing '
+          '"package:built_value/built_value.dart".');
```
manuelF: Maybe add a reason for it?

built_value_generator/lib/src/value_source_class.dart
davidmorgan: Thanks! I already found and fixed that on a different revision at least once. Now added a test for it like I should have done in the first place...

built_value_generator/lib/src/value_source_class.dart
davidmorgan: Done: 'it prevents the generated code from finding helper methods'

PR #369, opened 2018-03-24
Allow quiver 0.30

cbracken:
Increase allowed quiver dependency range to >=0.21.0 <0.30.0.

---
This change is [Reviewable](https://reviewable.io/reviews/google/built_value.dart/369)

### DIFF
--- a/benchmark/pubspec.yaml
+++ b/benchmark/pubspec.yaml
@@ -17,7 +17,7 @@ dependencies:
 dev_dependencies:
   build_runner: ^0.7.9
   built_value_generator: ^5.2.1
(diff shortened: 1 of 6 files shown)


cbracken:
quiver has a recent dev_dependency on build_runner, which depends on code_builder, which depends on (built_value, built_collection) which depends on quiver. 

As a result quiver currently has itself as a dev_dependency and can't be released without bumping the allowable version here.

/cc @matanlurey

dave26199:
Thanks. Will merge and push on Monday.

davidmorgan:
Thanks.

Travis is red because of pre-existing issues due to Dart 2 changes; fixed in separate PRs.

PR #377, opened 2018-03-28
Add builder option for header.

### LINKED ISSUE #374: Add options to builder
like https://github.com/dart-lang/json_serializable/blob/master/example/build.yaml

To add // ignore_for_file: ... or similar

...

Looks pretty easy to do, basically this line here:

https://github.com/dart-lang/json_serializable/blob/master/json_serializable/lib/builder.dart#L10

...

headers and additionalHeaders would be nice in case one only wants to add to default headers instead of replacing, but it's your call if you think this is worth it.

davidmorgan:
Fixes #374

---
This change is [Reviewable](https://reviewable.io/reviews/google/built_value.dart/377)

### DIFF
--- a/example/build.yaml
+++ b/example/build.yaml
@@ -1,0 +1,9 @@
+# This file is only needed if you want to tweak the built_value generator
+# options. The only available option is "header", which sets the header of
+# each generated file.
+targets:
+  $default:
+    builders:
+      built_value_generator|built_value:
+        options:
+          header: // This is generated code; please do not modify by hand.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+# 5.2.3 (not released)
+
+- Add support for setting the generated header in a `build.yaml` file in your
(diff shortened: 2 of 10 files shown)


example/build.yaml
```diff
@@ -0,0 +1,9 @@
+# This file is only needed if you want to tweak the built_value generator
+# options. The only available option is "header", which sets the header of
+# each generated file.
+targets:
+  $default:
+    builders:
+      built_value_generator|built_value:
+        options:
+          header: // This is generated code; please do not modify by hand.
```
zoechi: I think a multiline example like shown in https://github.com/dart-lang/json_serializable/blob/master/example/build.yaml would be better
I think it's more common and I didn't know before I saw that that YAML allows to do it this way

davidmorgan: Thanks for the suggestion. When I looked it up

https://yaml-multiline.info/

I found it's pretty complicated. I'll add a link to that page rather than an example.

zoechi: great

PR #419, opened 2018-05-24
Improve error message when deserialization fails.

### LINKED ISSUE #408: More Detailed Deserialization Error Messages
I'm working with a legacy API that quite honestly has odd structures. I'm looking at the backend Java code and trying to determine what to make my built_value models look like... however, I don't always get the types right. When I run a test on the data I will often get an error like

"String is not of type int"

But it'd be far more helpful to get something like "String get fieldName does not match {value} of type int".

Is this possible?

davidmorgan:
Fixes #408

---
This change is [Reviewable](https://reviewable.io/reviews/google/built_value.dart/419)

### DIFF
--- a/end_to_end_test/test/values_serializer_test.dart
+++ b/end_to_end_test/test/values_serializer_test.dart
@@ -5,6 +5,7 @@
 import 'package:built_value/serializer.dart';
 import 'package:built_value/standard_json_plugin.dart';
 import 'package:end_to_end_test/enums.dart';
+import 'package:end_to_end_test/errors_matchers.dart';
 import 'package:end_to_end_test/serializers.dart';
 import 'package:end_to_end_test/values.dart';
 import 'package:fixnum/fixnum.dart';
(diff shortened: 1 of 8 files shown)


end_to_end_test/test/values_serializer_test.dart
```diff
@@ -60,6 +61,33 @@ void main() {
...
+        'validatedValue',
+        [
+          'anInt',
+          'foo',
+        ],
+      ];
+      expect(
+          () => serializers.deserialize(incorrectSerialized),
+          throwsA(isErrorContaining(
+              "Deserializing '[CompoundValue, simpleValue, "
```
manuelF: Maybe missing some new lines? Error looks too collapsed to make sense of it IMO.

davidmorgan: I think it reads okay as a single line. I agree it's not terribly clear mashed into separate lines like this for the test :)

Thanks.

PR #439, opened 2018-06-12
Fix Intellij code template

BlackHC:
Needs $$ to escape $.

---
This change is [Reviewable](https://reviewable.io/reviews/google/built_value.dart/439)


davidmorgan:
Thanks! Someone changed it the other way; this time I actually checked which one works :)

BTW, you can use an analyzer plugin now; take a look at the `analysis_options.yaml` file for the project.

PR #464, opened 2018-07-24
Update to source_gen 0.9.0

natebosch:
Allows using `built_value` along side other `source_gen` based
generators that output to `.g.dart` part file. This is a breaking change
because the "header" can no longer be supported.

- Switch to `SharedPartBuilder` and add the required part ID argument.
- Move builds to the cache and update the new generated extension.
- Apply the `source_gen|combining_builder` to merge part files back into
  a `.g.dart` file.
- Remove the example `build.yaml` since it was only setting a `header`.
- Update test cases to keep using it as if it was a standalone part
  builder.

---
This change is [Reviewable](https://reviewable.io/reviews/google/built_value.dart/464)


built_value_generator/pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: built_value_generator
```
davidmorgan: Please run tool/use_local_deps and update the PR; it will update all the yaml files to use the unreleased version, so the presubmit results are meaningful.

natebosch: ran the tool and regenerated the example files.

davidmorgan:
Thanks!

zoechi:
What can `header` not be supported anymore?
Is there a replacement?
I usually depend on it to silence the analyzer when I enable new released linter rules.

natebosch:
@zoechi - I filed an issue with the reasoning and a potential solution at https://github.com/dart-lang/source_gen/issues/370

PR #475, opened 2018-08-08
Avoid dynamic in generated operator==

lexaknyazev:
Since positive `is` check enables type inference in analyzer and friends, `dynamic` could be avoided in the generated implementation of `operator==`.

---
This change is [Reviewable](https://reviewable.io/reviews/google/built_value.dart/475)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+# 6.0.1 (unreleased)
+
+- Improve generation for `operator==`, don't use `dynamic`.
+
 # 6.0.0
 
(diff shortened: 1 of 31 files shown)


lexaknyazev:
Done. 
For some reason, `dartfmt` altered formatting in `built_value_generator/lib/src/serializer_source_class.dart`.

davidmorgan:
Thanks!

dartfmt occasionally changes, so we do occasionally see extra reformats.

davidmorgan:
(Normally I try to publish changes within a couple of days, but I'll be on vacation for a week. So, it'll probably be a couple of weeks! Sorry about that.)

PR #505, opened 2018-09-17
Add readme examples for nullable and BuiltValueField

srsudar:
This adds examples for `@nullable` and `@BuiltValueField` to a new section on the README. As a newcomer to both dart and `built_value`, I would have appreciated a basic example when landing at the repo.

Dave mentioned on #144 that including this as an example might make sense:

> It would make sense to explain this and the other three related annotations on the main README file, I think.

I am not sure which other three those are, nor how to use the other ones I did see there, so I only included the two annotations mentioned above.

**Additionally, although this works for my particular use case, I would feel better about someone more familiar with the library double checking that my example makes sense before merging.**

---
This change is [Reviewable](https://reviewable.io/reviews/google/built_value.dart/505)

### DIFF
--- a/README.md
+++ b/README.md
@@ -213,6 +213,8 @@ import 'package:built_collection/built_collection.dart';
 part 'person.g.dart';
 
 abstract class Person implements Built<Person, PersonBuilder> {
+  static Serializer<Person> get serializer => _$personSerializer;
+
   // Can never be null.
   int get id;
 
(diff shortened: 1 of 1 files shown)


davidmorgan:
Very nice, thanks!

Could you please move the `serializer` getter to the topic of the class? That way it's consistent with the examples  https://github.com/google/built_value.dart/blob/master/example/lib/values.dart

PR #511, opened 2018-09-26
Modify pubspec to allow build package of 1.0 and greater

efortuna:
Previously built_value_generator relied on build ^0.0.12, and the build package is 1.0. I don't see any reason why I shouldn't allow 1.0, so I upped the version constraint.

Modify pubspec to allow build package of 1.0 and greater (up to next potentially breaking change of 2.0)

---
This change is [Reviewable](https://reviewable.io/reviews/google/built_value.dart/511)


natebosch:
@davidmorgan - we're upgrading the rest of the ecosystem too. Should be safe to allow the range.

PR #530, opened 2018-11-07
Update for analyzer 0.33.3.

davidmorgan:
This change is [Reviewable](https://reviewable.io/reviews/google/built_value.dart/530)

### DIFF
--- a/built_value_generator/lib/src/serializer_source_class.dart
+++ b/built_value_generator/lib/src/serializer_source_class.dart
@@ -74,12 +74,7 @@ abstract class SerializerSourceClass
 
   @memoized
   BuiltList<String> get genericParameters =>
-      new BuiltList<String>(element.typeParameters.map((e) {
-        final result = parsedLibrary.getElementDeclaration(e).node.toSource();
-        // Strip off any bounds.
-        final index = result.indexOf(' extends ');
-        return index == -1 ? result : result.substring(0, index);
-      }));
+      new BuiltList<String>(element.typeParameters.map((e) => e.name));
 
   @memoized
   BuiltList<String> get genericBounds =>
--- a/built_value_generator/lib/src/serializer_source_class.dart
+++ b/built_value_generator/lib/src/serializer_source_class.dart
@@ -4,8 +4,10 @@
 
 library built_value_generator.source_class;
 
+import 'package:analyzer/dart/analysis/results.dart';
 import 'package:analyzer/dart/element/element.dart';
 import 'package:analyzer/dart/element/type.dart';
+import 'package:analyzer/src/dart/analysis/results.dart'; // ignore: implementation_imports
 import 'package:built_collection/built_collection.dart';
 import 'package:built_value/built_value.dart';
 import 'package:built_value_generator/src/enum_source_class.dart';
@@ -29,13 +31,19 @@ abstract class SerializerSourceClass
               element.library.getType(element.displayName + 'Builder'));
   SerializerSourceClass._();
 
+  @memoized
+  ParsedLibraryResult get parsedLibrary =>
+      // ignore: deprecated_member_use
+      ParsedLibraryResultImpl.tmp(element.library);
+
   // TODO(davidmorgan): share common code in a nicer way.
   @memoized
   BuiltValue get builtValueSettings => new ValueSourceClass(element).settings;
 
   // TODO(davidmorgan): share common code in a nicer way.
   @memoized
-  BuiltValueEnum get enumClassSettings => new EnumSourceClass(element).settings;
+  BuiltValueEnum get enumClassSettings =>
+      new EnumSourceClass(parsedLibrary, element).settings;
 
   @memoized
   String get name => element.name;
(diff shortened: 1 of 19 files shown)


built_value_generator/lib/src/serializer_source_class.dart
```diff
@@ -57,12 +65,21 @@ abstract class SerializerSourceClass
...
+        '';
   }
 
   @memoized
-  BuiltList<String> get genericParameters => new BuiltList<String>(
-      element.typeParameters.map((e) => e.computeNode().toString()));
+  BuiltList<String> get genericParameters =>
+      new BuiltList<String>(element.typeParameters.map((e) {
+        final result = parsedLibrary.getElementDeclaration(e).node.toSource();
+        // Strip off any bounds.
```
davidmorgan: The original version you had gave different results, e.g. `T extends num` where previously I got `T`. So I just strip off any `extends` to get the same value.

scheglov: Ah, I guess I missed this place.
There was one more somewhere else.
You don't need to use code / string manipulations.

Something like
```dart
TypeParameter node = parsedLibrary.getElementDeclaration(e).node;
return node.name.name;
```

Or, even better, just
```
(e) => e.name;
```

I.e. if you need just name, get it from the element itself.

built_value_generator/lib/src/value_source_class.dart
```diff
@@ -109,9 +116,11 @@ abstract class ValueSourceClass
...
 
   @memoized
-  BuiltList<String> get genericParameters =>
-      new BuiltList<String>(element.typeParameters
-          .map((element) => element.computeNode().toString()));
+  BuiltList<String> get genericParameters => new BuiltList<String>(
+      element.typeParameters.map((element) => element.name));
+//  BuiltList<String> get genericParameters =>
+//      new BuiltList<String>(element.typeParameters.map((element) =>
+//          parsedLibrary.getElementDeclaration(element).node.toSource()));
```
scheglov: Ah, here is another case like below.
Remove commented out previous implementation?

built_value_generator/lib/src/value_source_class.dart
```diff
@@ -29,6 +31,11 @@ abstract class ValueSourceClass
       new _$ValueSourceClass._(element: element);
   ValueSourceClass._();
 
+  @memoized
+  ParsedLibraryResult get parsedLibrary =>
+      // ignore: deprecated_member_use
+      ParsedLibraryResultImpl.tmp(element.library);
```
scheglov: This is different from what was proposed in my internal CL.
Semantically it is the same, but performance is not the same.

Here you create a new `ParsedLibraryResult` instance every time, which includes parsing all library files again. This is a non-trivial amount of work. If you want it to be more efficient, create `ParsedLibraryResult` and pass it around.

built_value_generator/pubspec.yaml
```diff
@@ -11,7 +11,7 @@ environment:
   sdk: '>=2.0.0-dev <3.0.0'
 
 dependencies:
-  analyzer: ^0.33.1
+  analyzer: ^0.33.3
```
natebosch: @scheglov - Is the analyzer team prepared to support this API throughout the `0.33.x` version of Analyzer?

scheglov: @stereotype441 I know that there is a branch for analyzer `0.33.x` somewhere, but I was not able to find it on GitHub. IIRC we put this temporary non-API `ParsedLibraryResultImpl` and `ResolvedLibraryResultImpl` in this branch.

davidmorgan: @natebosch @scheglov Good point. I should pin this to exactly 0.33.3, I think?

stereotype441: Yes, we'll continue supporting this backported API for the rest of `0.33.x`, so it shouldn't be necessary to pin to exactly 0.33.3.  In fact, it would be better not to pin, because if we discover the need to do additional backporting to support other clients, we'll have to release `0.33.4`, and if/when we do that, we don't want to have to modify built_value to allow it.

@scheglov The branch is called "analyzer-0.33".  It's not mirrored to github, but it exists in googlesource.com, so assuming you clone from https://dart.googlesource.com/sdk.git, you should see the branch there.

davidmorgan: Sounds good, thanks. I'll aim to land this tomorrow.

built_value_generator/lib/src/serializer_source_class.dart
davidmorgan: That worked, thanks!

built_value_generator/lib/src/value_source_class.dart
davidmorgan: Done, thanks.

built_value_generator/lib/src/value_source_class.dart
davidmorgan: Thanks. Note that `@memoized` means it's _somewhat_ cached; it's only fetched once per containing class. This does mean we fetch more often. But it's neater ;)

I checked performance on the biggest built_value-containing package we have--82 classes--and there was no measurable slowdown. If anything this is a little faster than what we had before.

Once we're off the final solution I'll see if there's any further speedup to be had.

Thanks!

PR #553, opened 2018-12-11
Add an option to `@BuiltValue` to generate comparable builders.

### LINKED ISSUE #540: Add an option to make builders comparable
Some people want to compare builders. This is a lot of code to generate for people who don't want it, so make it an option.

It should be required to also turn off nested builders, because otherwise collections won't compare correctly.

davidmorgan:
Fixes #540

---
This change is [Reviewable](https://reviewable.io/reviews/google/built_value.dart/553)

### DIFF
--- a/built_value_generator/lib/src/value_source_class.dart
+++ b/built_value_generator/lib/src/value_source_class.dart
@@ -109,10 +109,16 @@ abstract class ValueSourceClass
         .where((value) => value?.type?.displayName == 'BuiltValue');
     if (annotations.isEmpty) return const BuiltValue();
     final annotation = annotations.single;
+    // If a field does not exist, that means an old `built_value` version; use
+    // the default.
     return new BuiltValue(
-        instantiable: annotation.getField('instantiable').toBoolValue(),
-        nestedBuilders: annotation.getField('nestedBuilders').toBoolValue(),
-        wireName: annotation.getField('wireName').toStringValue());
+        comparableBuilders:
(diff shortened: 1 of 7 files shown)


end_to_end_test/lib/values.g.dart
```diff
@@ -1163,6 +1222,118 @@ class CompoundValueNoNestingBuilder
...
+  bool operator ==(Object other) {
+    if (identical(other, this)) return true;
+    return other is CompoundValueComparableBuildersBuilder &&
+        simpleValue == other.simpleValue &&
+        validatedValue == other.validatedValue;
+  }
+
+  @override
+  int get hashCode {
+    return $jf($jc($jc(0, simpleValue.hashCode), validatedValue.hashCode));
```
emmanuel-p: Same hashcode as the Built, but not equal.
Should it have a different seed?

built_value_generator/lib/src/value_source_class.dart
```diff
@@ -299,6 +306,18 @@ abstract class ValueSourceClass
     }
   }
 
+  Iterable<GeneratorError> _checkSettings() {
+    if (settings.comparableBuilders && settings.nestedBuilders) {
```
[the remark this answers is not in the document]
davidmorgan: Done.

end_to_end_test/lib/values.g.dart
davidmorgan: Done, and added unit test.

Issue #559, opened 2018-12-30
Underscore in class name triggers inconsistent behaviour between error message and generated source


fibbers:
First of all, I know it's not good practice to have underscores in class names 😃 

If I have this code:
```dart
import 'package:built_value/built_value.dart';
import 'package:built_value/serializer.dart';

part 'book_author.g.dart';

abstract class Book_Author implements Built<Book_Author, Book_AuthorBuilder> {
  static Serializer<Book_Author> get serializer => _$book_AuthorSerializer;

  String get id;
  String get displayName;

  Book_Author._();
  factory Book_Author([updates(Book_AuthorBuilder b)]) = _$Book_Author;
}
```
the compiler succeeds, but generates source where the Serializer is being generated as:
```dart
Serializer<Book_Author> _$bookAuthorSerializer = new _$Book_AuthorSerializer();
```
instead of `_$book_AuthorSerializer`.

If I change my declaration to follow the generated source, I can't generate the source, since the build_runner now errors out:
```
1. Declare Book_Author.serializer as: static Serializer<Book_Author> get serializer => _$book_AuthorSerializer; got static Serializer<Book_Author> get serializer => _$bookAuthorSerializer;
```
So the error message seems inconsistent with the generated source.

PR #560, opened 2018-12-30
Fix handling of underscores in class name and serializer-declaration

fibbers:
I thought I'd create my first Google-PR in 2018 😄 so a first attempt to fix my own issue: #559.

I chose to follow the naming convention of the error message, since it stays as close to the source name as possible, the code that gives back the error-message was added more recently (with maybe newer insights), and a nice side effect is that it offers the opportunity to cleanup some code: `_toCamelCase`.

I also wanted to add a test, so I'm hoping I found the right place and tested the right things.

---
This change is [Reviewable](https://reviewable.io/reviews/google/built_value.dart/560)


dave26199:
Thanks! Will take a look when back from vacation in a week.

davidmorgan:
Hmm, that was quite complicated logic we had there. I wonder why. Maybe the end to end tests will show :)

To run the end to end tests, the packages need to refer to local versions instead of the ones on pub. I've done that here, please pull/rebase your PR

https://github.com/google/built_value.dart/pull/566

you can then locally run tool/presubmit to do the presubmit; on github travis should do it.

Note that changes to the generator can cause problems with package:build, since the generator runs on itself ... if on running tool/presubmit you hit generator errors you may need to run

```
cd built_value_generator
dart --snapshot=$HOME/bv_snapshot .dart_tool/build/entrypoint/build.dart; cp ~/bv_snapshot ~/bv_snapshot2; dart ~/bv_snapshot2 build
```

--if it asks you whether to delete files, choose 'delete', they'll be regenerated.

Once you have this working, the test would be better added under `end_to_end_test`, by adding a class to  `end_to_end_test/values.dart` with underscores in the name; and a test that uses the serializer to `test/values_serializer_test.dart`.

Thanks!

davidmorgan:
Whoops, looks like the end to end tests are currently failing due to an SDK change

https://travis-ci.org/google/built_value.dart/builds/477431474?utm_medium=notification&utm_source=email

--should be a quick fix, I'll try to take a look.

davidmorgan:
This fixes the presubmit, so it should be useful again

https://github.com/google/built_value.dart/pull/567

fibbers:
I moved the test-code to `end_to_end_test` as you suggested, but since the code is now actually being generated into `values.g.dart` there are now linting errors (`non_constant_identifier_names`/`camel_case_types`):
```sh
lint • Name non-constant identifiers using lowerCamelCase at lib/values.g.dart:49:5 • non_constant_identifier_names
lint • Name types using UpperCamelCase at lib/values.g.dart:976:7 • camel_case_types
lint • Name types using UpperCamelCase at lib/values.g.dart:3684:7 • camel_case_types
lint • Name types using UpperCamelCase at lib/values.g.dart:3727:7 • camel_case_types
```

I'm not sure how I should handle this. I [ignored linting errors in `values.dart`](https://github.com/fibbers/built_value.dart/commit/42589e755f500ef7ad119f301278def6e8c383bb#diff-0495adf5d2e9dbde8ab38eae465d6ad1R477):
```dart
// ignore: camel_case_types
abstract class Value_With_Underscore_In_Name
```
but the generated code doesn't have these ignore-comments.

I can think of a few ways to solve this:
- don't lint the `.g.dart` files, however I don't think this is preferable, since people using this project probably also don't want linting errors in `.g.dart` files, hence they should still show up
- manually modifying the `.g.dart` file... not a good solution either
- generating the ignore-comments... not trivial, since the generated code could need different comments than the source (in my case: the source needs `camel_case_types`, but the generated needs `non_constant_identifier_names`+`camel_case_types`)
- maybe move the test into its own source+generated files where the generated one is ignored in linting
- ...something else?

What do you think?

Thanks for taking a look at this btw!

fibbers:
In my own projects where I use this library, I already modified my code to not have underscores in the classnames and I don't think that's too much of a problem (maybe even the opposite).

I'm thinking this functionality might need too many changes with too low of a payoff, so while it was fun working on this, in the spirit of "kill your darlings" I'll close this PR :-)
(If you _do_ think we should go further with this, I'm open to suggestions)

PR #584, opened 2019-02-05
Add ignored linter conflicting with flutter defaults

atonem:
The generated files created seems to not include `always_specify_types` among the generated ignored linters. This causes problems with the default flutter `analysis_options.yaml` and results in a pretty unreadable analyzer output.

---
This change is [Reviewable](https://reviewable.io/reviews/google/built_value.dart/584)

### DIFF
--- a/built_value_generator/lib/built_value_generator.dart
+++ b/built_value_generator/lib/built_value_generator.dart
@@ -64,6 +64,7 @@ class BuiltValueGenerator extends Generator {
           '\n'
           '// ignore_for_file: '
           'always_put_control_body_on_new_line,'
+          'always_specify_types'
           'annotate_overrides,'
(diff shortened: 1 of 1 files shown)


built_value_generator/lib/built_value_generator.dart
```diff
@@ -64,6 +64,7 @@ class BuiltValueGenerator extends Generator {
           '\n'
           '// ignore_for_file: '
           'always_put_control_body_on_new_line,'
+          'always_specify_types'
```
davidmorgan: Please add a comma at the end.

davidmorgan:
Thanks!

It may be a little while before I can do a release, I'm just about to go on vacation for a week; sorry about that.

PR #610, opened 2019-03-22
Removes unnecessary "new" and "const" from example projects

rkinabhi:
Reduce verbosity by removing unnecessary "new" and "const" from 

- [_example_](https://github.com/google/built_value.dart/tree/master/example)
- [_chat_example_](https://github.com/google/built_value.dart/tree/master/chat_example)

---
This change is [Reviewable](https://reviewable.io/reviews/google/built_value.dart/610)


davidmorgan:
Thanks for the cleanup :)

PR #612, opened 2019-03-23
Removes unnecessary "new" and "const".

rkinabhi:
This PR reduces verbosity by removing unnecessary "new" and "const" from 

- [_benchmark_](https://github.com/google/built_value.dart/tree/master/benchmark)
- [_built_value_](https://github.com/google/built_value.dart/tree/master/built_value)
- [_built_value_generator_](https://github.com/google/built_value.dart/tree/master/built_value_generator)
- [_built_value_test_](https://github.com/google/built_value.dart/tree/master/built_value_test)
- [_end_to_end_test_](https://github.com/google/built_value.dart/tree/master/end_to_end_test)

PS: I would also like to work on a patch that would make the library generate code without linting errors such as the [_unnecessary_new_](http://dart-lang.github.io/linter/lints/unnecessary_new.html) and [_unnecessary_const_](http://dart-lang.github.io/linter/lints/unnecessary_const.html). If this patch would be helpful, let me know if i can make those changes as part of this current PR or open a new PR for the patch. 😄

---
This change is [Reviewable](https://reviewable.io/reviews/google/built_value.dart/612)

### DIFF
--- a/benchmark/lib/benchmark.dart
+++ b/benchmark/lib/benchmark.dart
@@ -11,7 +11,7 @@ void benchmark() {
 }
 
 void benchmarkHashCode() {
-  final value = new SimpleValue((b) => b
+  final value = SimpleValue((b) => b
     ..anInt = 0
     ..aString = 'zero');
   _benchmark('hashCode', () => value.hashCode);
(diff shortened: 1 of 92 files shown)


davidmorgan:
This one is a bit trickier, I'm afraid :)

We have some legacy code that needs to be dart one compatible and is using `built_value`. So the generated code, at least, needs to keep using `new` and `const`.

I'm not 100% sure if it will be a breaking change to remove `new` and `const` from `built_value` and `built_value_generator`; I'll need to check.

So here's my suggestion:

- For this PR, leave `built_value` and `built_value_generator` alone;
- Split `built_value` and `built_value_generator` changes into a separate PR; I'll check whether they break our legacy code
- For the generated code itself, unfortunately we can't currently do better than using `ignore_for_file` as we do now

Thanks :)

davidmorgan:
Thanks!

PR #616, opened 2019-03-27
Extendable nullable annotations

rmawatson:
When using another generator with built_value, this enables other other generator's annotations to extend built_value's `nullable` using `NullableAnnotation`, 

```
@nullable
@OtherGeneratorAnnotation
get int field;
```

Where `OtherGenerator` requires fields marked `@OtherGeneratorAnnotation` are also `@nullable`, this removes the need to have to explicitly annotate with `@nullable `and `@OtherGeneratorAnnotation` (and potentially forget to add `@nullable`) by having `OtherGeneratorAnnotation` extend `NullableAnnotation`.

```
@OtherGeneratorAnnotation
get int field;
```

---
This change is [Reviewable](https://reviewable.io/reviews/google/built_value.dart/616)

### DIFF
--- a/built_value/lib/built_value.dart
+++ b/built_value/lib/built_value.dart
@@ -103,10 +103,17 @@ class BuiltValue {
       this.wireName});
 }
 
-/// Nullable annotation for Built Value fields.
+/// NullableAnnotation used by nullable
+///
+/// Can be extended/implemented by another annotation
+class NullableAnnotation {
+  const NullableAnnotation();
+}
+
+/// nullable annotation for Built Value fields.
 ///
 /// Fields marked with this annotation are allowed to be null.
-const String nullable = 'nullable';
+const NullableAnnotation nullable = NullableAnnotation();
 
 /// Optionally, annotate a Built Value field with this to specify settings.
 /// This is only needed for advanced use.
--- a/built_value_generator/lib/src/metadata.dart
+++ b/built_value_generator/lib/src/metadata.dart
@@ -23,6 +23,13 @@ String metadataToStringValue(ElementAnnotation annotation) {
   return value.toStringValue();
 }
 
+/// Gets the `Type` value of an annotation. Throws a descriptive
+/// [InvalidGenerationSourceError] if the annotation can't be resolved.
+DartObject metadataToObjectValue(ElementAnnotation annotation) {
+  final value = getConstantValueFromAnnotation(annotation);
+  return value;
+}
+
 /// Gets a field from an annotation. Throws a descriptive
 /// [InvalidGenerationSourceError] if the annotation can't be resolved.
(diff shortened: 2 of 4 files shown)


davidmorgan:
Sorry for the slow response.

This would be a breaking change to `built_value`; older `built_value_generator` will not recognize the new annotations, and new `built_value_generator ` will not recognize the old annotations. I'd prefer to avoid that if possible.

Also, when NNBD arrives, we won't need the `@nullable` annotation any more; it will be part of the type system, i.e. you'll write `int?` instead of `int`. That should be this year. So maybe it makes more sense to just wait until then.

I'd be interested to hear more about what this is for, though. If there's a simple way to support the use case then I'm all for it.

Thanks!

rmawatson:
Hi David. We specifically wanted this to be able to annotate with our own generator annotation, where built_value_generator is also running on the same file & same class definition. We required that anything annotated with our annotation was also annotated with built_value @nullable. The neatest way seemed to be to have our annotation implement the PR's nullable class, so our annotation implicitly implies built_values @nullable Obviously annotating the get with both @nullable and @ourannotation works, but forgetting @nullable would break our generated code. We've decided to abandon built_value and instead just implement our own equivalent in the our generator, which is a bit of a shame.

In regards to this being a breaking change. I was under the impression that you would upgrade built_value and built_value_generator in lockstep?

davidmorgan:
It sounds like your generator is not open source? That might make it a bit easier to maintain. FWIW, if you make the `toString` of your annotation return `nullable`, I think that will work today :)

`built_value` and `built_value_generator` used to be pinned to the same version, but this was making it harder for people to use because of other dependency issues; so currently `built_value_generator` only requires `built_value 5.5.5`.

rmawatson:
Thanks good to know regarding the dependency issues.

With regards to toString() on the annotation, toStringValue() that is called on the DartObject (by your original code) appears to only return a String where the underlying type is String. when the type is SomeAnnotation, this doesn't consider any `toString()` function that may be available on the class definition. At least from what i've tested this always returns null.

rmawatson:
Just wanted to add that the PR style does seem more inline with how the dart core declares various built in annotations. https://github.com/dart-lang/sdk/blob/master/sdk/lib/core/annotations.dart

davidmorgan:
Ah, you're right, `toStringValue` is only for strings, it won't work.

We could make it check for both the old and new annotation. But, that still leaves a failure case if new `built_value` is used with old `built_value_generator`. So it would still need a version bump to `built_value`, which I would prefer not to do.

The only other thing I can think is that you could maintain a local fork of `built_value_generator` with this modification. Given that NNBD is intended to land this year, the need for the `@nullable` annotation should go away, so it will be a temporary workaround.

davidmorgan:
Closing this--we'll need to figure out something backwards compatible to make this change.

PR #627, opened 2019-04-05
Add `@BuiltValueEnumConst(fallback: true)`.

### LINKED ISSUE #617: Forward-compatible enums
As an example, I have an enum class like:

```dart
class Fruits extends EnumClass {
  const Fruits._(String name) : super(name);
  static Serializer<Fruits> get serializer => _$fruitsSerializer;

  @BuiltValueEnumConst(wireName: 'APPLE')
  static const Fruits apple = _$apple;

  @BuiltValueEnumConst(wireName: 'ORANGE')
  static const Fruits orange = _$orange;

  @BuiltValueEnumConst(wireName: 'BANANA')
  static const Fruits banana = _$banana;

  static BuiltSet<Fruits> get values => _$values;
  static Fruits valueOf(String name) => _$valueOf(name);
}
```

I'm using this enum to deserialize a string that's coming from an API. Now, someone else on the API team updates things, and now the API is returning another fruit, `kiwi`. Immediately my app crashes because this enum cannot deserialize `kiwi`. Ideally, I would like to deserialize any "other" value into an `other` enum value. Or simply deserializing to `null` would also work. Is there any way to handle this kind of thing?

davidmorgan:
Fixes #617 

Also reverts a couple of 'final' -> 'var' changes that were to the generated output rather than to the generator/tests.

---
This change is [Reviewable](https://reviewable.io/reviews/google/built_value.dart/627)


davidmorgan:
Thanks!

PR #629, opened 2019-04-05
Stop generating builder factories for non-built generic fields.

### LINKED ISSUE #608: SerializersFor assumes a Builder for custom types
A basic reproduction case (more details can be added if needed):

```dart
// example.dart

// A built-class.
abstract class Bar implements Built<Bar, BarBuilder> {
  Drink<Fizzy> get drink;
}

// Not a built-class.
class Drink<T> { /* ... */ }

@SerializersFor([
  Bar,
])
final Serializers mySerializers = _$mySerializers;
```

... tries to generate:

```dart
// example.g.dart

// ...
      ..addBuilderFactory(
          const FullType(Drink, const [const FullType(Fizzy)]),
          () => new DrinkBuilder<Fizzy>())
// ...
```

This is not valid code, as `Drink` is not a `Built` and does not have a `Builder`.  I can get "around" this for now by adding essentially useless code to prevent a build error:

```dart
class DrinkBuilder<T> {
  DrinkBuilder() {
    assert(false, 'Here to get around a bug, sorry!');
  }
}
```

davidmorgan:
Fixes #608

---
This change is [Reviewable](https://reviewable.io/reviews/google/built_value.dart/629)


davidmorgan:
Thanks. Yes, I needed the 'var->final' revert so the generated output doesn't change. (I neglected to regenerate it when I submitted that CL, apparently, which is why we didn't notice it changed). I should have split that into a separate PR, sorry for that.

PR #631, opened 2019-04-05
Add `@BuiltValueSerializer(serializeNulls)`.

### LINKED ISSUE #604: null values are stripped when serializing
Hello, thanks for a wonderful plugin. It's making our boilerplate related headacres go away.

However we're running into a minor issue serializing data into a Firestore compatible Map. The generated serializers are stripping out nulls (which makes sense in a JSON context).

The offending code is the `if` that is generated in the `.g.dart` files, for example:
```
if (object.color != null) {
      result
        ..add('color')
        ..add(serializers.serialize(object.color,
            specifiedType: const FullType(String)));
}
```

If we comment it out, our resulting Map has keys with nulls, as desired.

Any way to make built_value not generate that `if` check?

davidmorgan:
Fixes #604


larssn:
You're doing a stellar job mr Morgan!

PR #636, opened 2019-04-16
Add an ISO8601 serializer for Duration

jadengis:
I was using built_value to interface with a Spring server serving Durations in the ISO8601 format as described [here](https://en.wikipedia.org/wiki/ISO_8601#Durations). The microsecond Duration serializer was insufficient for handling this data.

I wrote a serializer in a in a similar vein to the already existing `Iso8601DateTimeSerializer` to handle these formatted durations.

This serializer does not support the full range of acceptable ISO8601 durations due to the limitations of the Dart `Duration` object (i.e. no year or week settings), but works for the range of values covered by
the `Duration` object.

This serializer is complete with tests similar to the ones for `Iso8601DateTimeSerializer`, and cover a lot of the edge cases, but I would be happy to add more if need be.

Please let me know if there is anything else you would like me to change.

### DIFF
--- a/built_value/lib/iso_8601_duration_serializer.dart
+++ b/built_value/lib/iso_8601_duration_serializer.dart
@@ -1,0 +1,96 @@
+import 'package:built_collection/built_collection.dart';
+import 'package:built_value/serializer.dart';
+
+/// Alternative serializer for [Duration].
+///
+/// Install this to use ISO8601 compatible format instead of the default
+/// (microseconds). Use [SerializersBuilder.add] to install it.
+///
+/// Note that this serializer is not 100% compatible with the ISO8601 format
+/// due to limitations of the [Duration] class, but is designed to produce and
+/// consume reasonable strings that match the standard.
+class Iso8601DurationSerializer extends PrimitiveSerializer<Duration> {
+  @override
+  Duration deserialize(Serializers serializers, Object serialized,
+      {FullType specifiedType = FullType.unspecified}) {
+    final String value = serialized;
+    return _parseDuration(value);
+  }
+
+  @override
+  Object serialize(Serializers serializers, Duration object,
+      {FullType specifiedType = FullType.unspecified}) {
+    return _writeIso8601Duration(object);
+  }
+
+  @override
+  Iterable<Type> get types => BuiltList(const [Duration]);
+
+  @override
+  String get wireName => 'Duration';
+
+  Duration _parseDuration(String value) {
+    final match = _parseFormat.firstMatch(value);
+    if (match != null) {
+      // Iterate through the capture groups to build the unit mappings
+      final unitMappings = <String, int>{};
+
+      // Start iterating at 1, because match[0] is the full match
+      for (var i = 1; i <= match.groupCount; i++) {
+        final group = match[i];
+        if (group == null) {
+          continue;
+        }
+        // Get all but last character in group
+        // RegExp ensures this must be an int
+        final value = int.parse(group.substring(0, group.length - 1));
+        // Get last character
+        final unit = group.substring(group.length - 1);
+        unitMappings[unit] = value;
+      }
+      return Duration(
+        days: unitMappings[_dayToken] ?? 0,
+        hours: unitMappings[_hourToken] ?? 0,
+        minutes: unitMappings[_minuteToken] ?? 0,
+        seconds: unitMappings[_secondToken] ?? 0,
+      );
+    } else {
+      throw FormatException("Invalid duration format", value);
+    }
+  }
+
+  String _writeIso8601Duration(Duration duration) {
+    if (duration == Duration.zero) {
+      // format for zero.
+      return 'PT0S';
+    }
+    final days = duration.inDays;
+    final hours = (duration - Duration(days: days)).inHours;
+    final minutes = (duration - Duration(days: days, hours: hours)).inMinutes;
+    final seconds =
+        (duration - Duration(days: days, hours: hours, minutes: minutes))
+            .inSeconds;
+    final buffer = StringBuffer(_durationToken)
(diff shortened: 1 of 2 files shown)


built_value/lib/iso_8601_duration_serializer.dart
```diff
@@ -0,0 +1,96 @@
...
+/// Note that this serializer is not 100% compatible with the ISO8601 format
+/// due to limitations of the [Duration] class, but is designed to produce and
+/// consume reasonable strings that match the standard.
+class Iso8601DurationSerializer extends PrimitiveSerializer<Duration> {
+  @override
+  Duration deserialize(Serializers serializers, Object serialized,
+      {FullType specifiedType = FullType.unspecified}) {
+    final String value = serialized;
+    return _parseDuration(value);
+  }
```
davidmorgan: You don't need `value`; this can just be a one liner method with =>.

built_value/lib/iso_8601_duration_serializer.dart
```diff
@@ -0,0 +1,96 @@
...
+
+  @override
+  Object serialize(Serializers serializers, Duration object,
+      {FullType specifiedType = FullType.unspecified}) {
+    return _writeIso8601Duration(object);
+  }
```
davidmorgan: Please use =>.

built_value/lib/iso_8601_duration_serializer.dart
```diff
@@ -0,0 +1,96 @@
...
+        unitMappings[unit] = value;
+      }
+      return Duration(
+        days: unitMappings[_dayToken] ?? 0,
+        hours: unitMappings[_hourToken] ?? 0,
+        minutes: unitMappings[_minuteToken] ?? 0,
+        seconds: unitMappings[_secondToken] ?? 0,
+      );
+    } else {
+      throw FormatException("Invalid duration format", value);
```
davidmorgan: Please move this up as a 'fail early' to reduce indentation.

if (match == null) throw FormatException(...).

Additionally, please use single quotes for strings.

built_value/lib/iso_8601_duration_serializer.dart
```diff
@@ -0,0 +1,96 @@
...
+  @override
+  Iterable<Type> get types => BuiltList(const [Duration]);
+
+  @override
+  String get wireName => 'Duration';
+
+  Duration _parseDuration(String value) {
+    final match = _parseFormat.firstMatch(value);
+    if (match != null) {
+      // Iterate through the capture groups to build the unit mappings
```
davidmorgan: https://www.dartlang.org/guides/language/effective-dart/documentation#do-format-comments-like-sentences

here and below.

built_value/lib/iso_8601_duration_serializer.dart
```diff
@@ -0,0 +1,96 @@
...
+      final unitMappings = <String, int>{};
+
+      // Start iterating at 1, because match[0] is the full match
+      for (var i = 1; i <= match.groupCount; i++) {
+        final group = match[i];
+        if (group == null) {
```
davidmorgan: This can be a single line if you like. Up to you.

if (group == null) continue;

built_value/lib/iso_8601_duration_serializer.dart
```diff
@@ -0,0 +1,96 @@
...
+        seconds: unitMappings[_secondToken] ?? 0,
+      );
+    } else {
+      throw FormatException("Invalid duration format", value);
+    }
+  }
+
+  String _writeIso8601Duration(Duration duration) {
+    if (duration == Duration.zero) {
+      // format for zero.
```
davidmorgan: I don't think this comment adds anything. Up to you.

built_value/lib/iso_8601_duration_serializer.dart
```diff
@@ -0,0 +1,96 @@
...
+            .inSeconds;
+    final buffer = StringBuffer(_durationToken)
+      ..write(days == 0 ? '' : '$days$_dayToken');
+    if (!(hours == 0 && minutes == 0 && seconds == 0)) {
+      buffer
+        ..write(_timeToken)
+        ..write(hours == 0 ? '' : '$hours$_hourToken')
+        ..write(minutes == 0 ? '' : '$minutes$_minuteToken')
+        ..write(seconds == 0 ? '' : '$seconds$_secondToken');
+    }
```
davidmorgan: Does the format support subsecond precision?

If not, it would probably be best to throw an exception if there is anything left over that can't be serialized, i.e. any sub-second data.

built_value/lib/iso_8601_duration_serializer.dart
```diff
@@ -0,0 +1,96 @@
...
+  static const _durationToken = 'P';
+  static const _dayToken = 'D';
+  static const _timeToken = 'T';
+  static const _hourToken = 'H';
+  static const _minuteToken = 'M';
+  static const _secondToken = 'S';
+
+  // Parse format for ISO8601
+  static final _parseFormat = RegExp(
+      '^P(?!\$)(0D|[1-9][0-9]*D)?(?:T(?!\$)(0H|[1-9][0-9]*H)?(0M|[1-9][0-9]*M)?(0S|[1-9][0-9]*S)?)?\$');
```
davidmorgan: Please split the string so this fits in 80 cols.

built_value/lib/iso_8601_duration_serializer.dart
jadengis: The format does not support the sub-second precision. I'll add the safety check.

jadengis:
@davidmorgan I addressed you comments an updated the CHANGELOG. Let me know if the new version number is okay.

davidmorgan:
Looks good, thanks!

PR #652, opened 2019-05-24
Stop making use of the `tmp` constructor for `ParsedLibraryResultImpl`.

stereotype441:
This was a temporary analyzer API; it has been deprecated and replaced
by a method in AnalysisSession.

### DIFF
--- a/built_value_generator/lib/src/enum_source_class.dart
+++ b/built_value_generator/lib/src/enum_source_class.dart
@@ -6,7 +6,6 @@ library built_value_generator.enum_source_class;
 
 import 'package:analyzer/dart/analysis/results.dart';
 import 'package:analyzer/dart/element/element.dart';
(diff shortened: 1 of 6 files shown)


stereotype441:
@davidmorgan

davidmorgan:
Thanks! Do you need me to release/import this any time soon?

stereotype441:
> Thanks! Do you need me to release/import this any time soon?

It needs to be rolled internally, but an external release isn't necessary.  If you wouldn't mind, that would be great!

davidmorgan:
@stereotype441 will do both, built_value is due a release anyway. Thanks :)

Edijae:
@stereotype441  which is the method and how do we use it to replace

`var parsedLibrary = ParsedLibraryResultImpl.tmp(element.library);`

Edijae:
@stereotype441  I got it

`var parsedLibrary = element.library.session.getParsedLibraryByElement(element.library);`

PR #662, opened 2019-06-11
Enable syntax highlighting in README code blocks

greglittlefield-wf:
README.md was using plain code blocks, which resulted in no syntax highlighting, making those snippets harder to read.

This PR adds languages to the blocks to enable highlighting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ code. Luckily, even this bit of boilerplate can be got automated using code
 snippets support in your favourite text editor. For example, in IntelliJ you
 can use following live template:
 
-```
+```dart
 abstract class $CLASS_NAME$ implements Built<$CLASS_NAME$, $CLASS_NAME$Builder> {
   $CLASS_NAME$._();
   factory $CLASS_NAME$([void Function($CLASS_NAME$Builder) updates]) = _$$$CLASS_NAME$;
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -188,7 +188,7 @@ While full, compiled examples are available in
 a common usage example is shown here. This example assumes that you are writing
 a client for a JSON API representing a person that looks like the following:
 
-```
+```js
```
greglittlefield-wf: I went with JS since this isn't valid JSON.

JSON:
```js
{
  id: 12345,
  age: 35,
  first_name: 'Jimmy',
  hobbies: ['jumping', 'basketball']
}
```

js:
```json
{
  id: 12345,
  age: 35,
  first_name: 'Jimmy',
  hobbies: ['jumping', 'basketball']
}
```

I can revert this back to unhighlighted if that is preferred.

davidmorgan: Thanks for the PR!

I guess to be valid JSON the keys should be strings? I think that was actually what was intended. (Certainly, the keys _do_ need to be strings). Does that make it happy with 'json'?

greglittlefield-wf: Sure thing!

Yup, I just pushed up a commit that switches it to JSON syntax/highlighting

davidmorgan:
Thanks :)

PR #688, opened 2019-07-29
Change analyzer_plugin version range.

scheglov:
We change `DartEditBuilder.writeOverride` in https://dart-review.googlesource.com/c/sdk/+/110941, as far as I can see, this does not affect your plugin. So, you can set your version range for "analyzer_plugin" higher.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+# 6.7.1
+
+- Bump version of `analyzer_plugin`.
+
 # 6.7.0
 
 - Generate code compatible with 'no raw types'.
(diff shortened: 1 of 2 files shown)


dave26199:
Sounds good, thanks.

Is anything blocked on getting this merged/published? I'm on vacation this week :)

scheglov:
There is nothing to do now, have fun on your vacation :-)

I just want to make sure that clients of `analyzer_plugin` know that they can use wider version range, so that if we do mode work to improve it, they get these improvements.

I don't have write access, so cannot land this PR myself, but it totally can wait until after your vacation.

PR #698, opened 2019-08-22
serializers: fix syntax in addBuilderFactory dartdoc

maxlapides:
The issue I'm fixing here is the second parameter passed to `addBuilderFactory` should be `() => ListBuilder<Foo>()` but currently it's missing the `()` after `<Foo>`. I also updated the syntax to remove the `new` keyword and formatted the code with dartfmt.


maxlapides:
@googlebot I signed it!

davidmorgan:
Thanks :)

PR #718, opened 2019-10-18
Use DartTypes.getName() instead of DartType.displayName

scheglov:
Stop using DartType.toString() for code generation.

We would like to change DartType.toString() to include nullability suffixes `?` and `*`. And it actually has never been a really good idea to use `toString()` or `DartType.displayName` for code generation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+# 6.7.2 (not yet published)
+
+- Replace analyzer's `DartType.displayName` with custom code generator.
+
 # 6.7.1
 
 - Fix codegen for custom builders when fields use types that have an import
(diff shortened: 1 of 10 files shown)


davidmorgan:
Thanks Konstantin! Just got back from vacation; I was planning anyway to do some `built_value` fixes and a release this week, so I'll try to take a look at this today or tomorrow. Any urgency from your side?

davidmorgan:
PR looks good, thanks!

So for the "any urgency" question: any urgency on an external release and/or pulling into google3? Thanks.

scheglov:
There is no hurry at this moment.

We need to make changes to presentation of types in DAS carefully in order to not expose nullability suffixes in libraries that are not oped-in. It will take some time to get to it, and implement.

davidmorgan:
Thanks. It'll likely get releases this week, anyway :)

PR #723, opened 2019-10-21
Fix `operator==` when a field is called `other`.

### LINKED ISSUE #713: Fields named "other" result in incorrect code generated.
Given a built value with a field named `other`:

```dart
abstract class Something implements Built<Something, SomethingBuilder> {
  factory Something([void Function(SomethingBuilder) updates]) = _$Something;

  Something._();

  static Serializer<Something> get serializer => _$somethingSerializer;

  String get other;
}
```

The following code is generated for operator `==`:

```dart
  @override
  bool operator ==(Object other) {
    if (identical(other, this)) return true;
    return other is Something && other == other.other;
  }
```
This incorrectly refers to the parameter `other` rather than the field `other`.

davidmorgan:
Fixes #713


built_value_generator/lib/src/value_source_class.dart
```diff
@@ -933,11 +933,13 @@ abstract class ValueSourceClass
...
     if (comparedFields.isNotEmpty) {
       result.writeln('&&');
-      result.writeln(comparedFields
-          .map((field) => field.isFunctionType
-              ? '${field.name} == _\$dynamicOther.${field.name}'
-              : '${field.name} == other.${field.name}')
-          .join('&&'));
+      result.writeln(comparedFields.map((field) {
+        var nameOrThisDotName =
+            field.name == 'other' ? 'other.${field.name}' : field.name;
```
iinozemtsev: Should it be `'this.'`?

iinozemtsev: I think it would be easier to read just like this, without string substitution: `field.name == 'other' ? 'this.other' : field.name`

davidmorgan: Oh man, that's embarrassing. Thanks. Done (both), and updated the test to test _not equal_ as well as _equal_.

PR #728, opened 2019-10-21
Make codegen trigger less eager.

### LINKED ISSUE #692: Using value type as Mocks for testing
I am using value types in my core code, but for testing purposes, I am mocking them - `MockValueTypeClass extends Mock implements ValueTypeClass {}`. With this in the test code, when I am trying to auto-generate the code, it shows error in the test code, which is obvious as the `MockValueTypeClass` is not following standards in which a value type class should be implemented and yet implementing (indirectly) the builder pattern. Is their a way to stop built-value generator from looking into the test code and just go through the `lib` directory ?

davidmorgan:
Fixes #692


davidmorgan:
Doesn't work, needs more thought.

PR #793, opened 2020-01-14
Fix generation for implicit-casts and implicit-dynamic.

### LINKED ISSUE #669: Cast in deserialize method of generated serializer fails strong mode linting with @BuiltValue(nestedBuilders: false)
I have implicit casts disabled in my `analysis_options.yaml`.

i.e.
``` yaml
analyzer:
  strong-mode:
    implicit-casts: false
    implicit-dynamic: false
```

I have the class below:

``` dart
@BuiltValue(nestedBuilders: false)
abstract class Example implements Built<Example, ExampleBuilder> {
  static Serializer<Example> get serializer => _$exampleSerializer;

  BuiltMap<String, JsonObject> get example;

  Example._();
  factory Example([void Function(ExampleBuilder) updates]) = _$Example;
}
```

I get the following generated serializer, which fails linting:
``` dart
class _$ExampleSerializer implements StructuredSerializer<Example> {
  @override
  final Iterable<Type> types = const [Example, _$Example];
  @override
  final String wireName = 'Example';

  @override
  Iterable serialize(Serializers serializers, Example object,
      {FullType specifiedType = FullType.unspecified}) {
    final result = <Object>[
      'example',
      serializers.serialize(object.example,
          specifiedType: const FullType(BuiltMap,
              const [const FullType(String), const FullType(JsonObject)])),
    ];

    return result;
  }

  @override
  Example deserialize(Serializers serializers, Iterable serialized,
      {FullType specifiedType = FullType.unspecified}) {
    final result = new ExampleBuilder();

    final iterator = serialized.iterator;
    while (iterator.moveNext()) {
      final key = iterator.current as String;
      iterator.moveNext();
      final dynamic value = iterator.current;
      switch (key) {
        case 'example':
          result.example = serializers.deserialize(value,
              specifiedType: const FullType(BuiltMap, const [
                const FullType(String),
                const FullType(JsonObject)
              ])) as BuiltMap;  // <---- This should be BuiltMap<String, JsonObject>
          break;
      }
    }

    return result.build();
  }
}
```

It's probably worth noting that it seems to only happen when using `BuiltMap`, `BuiltList` etc. If I have `Map<String, JsonObject> get example;` then the serializer will be generated with `... as Map<String, JsonObject>`.

And it is only an issue when using `@BuiltValue(nestedBuilders: false)`.

davidmorgan:
Fixes #669

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 - Make generated code comply with analyzer option `strict-raw-types`.
 - Allow `Serialiers` declaration to comply with `strict-raw-types`, or to
   continue to use raw types as before.
+- Make generated code comply with analyzer options `implicit-casts: false`
+  and `implicit-dynamic: false`.
 
 # 7.0.5
 
--- a/built_value_generator/lib/src/serializer_source_field.dart
+++ b/built_value_generator/lib/src/serializer_source_field.dart
@@ -192,25 +192,7 @@ abstract class SerializerSourceField
       String type, BuiltMap<String, String> classGenericBounds,
       {bool topLevel = true}) {
     var bareType = _getBareType(type);
-
-    // For built collections we can cast to the dynamic type when deserializing,
-    // instead of the actual generic type. This is because the `replace` method
-    // checks the generic type and copies if needed.
-    String generics;
-    if (topLevel && DartTypes.isBuiltCollectionTypeName(bareType)) {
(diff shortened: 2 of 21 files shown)


davidmorgan:
Whoops, looks like I have a bit more to fix. Will do so ...

davidmorgan:
It looks like the analyzer changed to not report these errors, which is why I can't see them locally. Filed an issue, will follow up when I have a response

https://github.com/dart-lang/sdk/issues/40129

PR #816, opened 2020-01-28
Example: replace version with `publish_to: none`

kevmoo:
Make it very clear this is not meant to be published!


davidmorgan:
Neat. Thanks!

PR #819, opened 2020-02-04
Remove the comment about @virtual

TastyPi:
`@virtual` is no longer needed and the field isn't annotated with it anyway.


davidmorgan:
Thanks!

PR #820, opened 2020-02-05
Use DartTypes.getName() instead of DartType.toString()

scheglov:
We did this some time ago, but because it is not enforced, probably another one crept in.

### DIFF
--- a/built_value_generator/lib/src/serializer_source_class.dart
+++ b/built_value_generator/lib/src/serializer_source_class.dart
@@ -111,7 +111,7 @@ abstract class SerializerSourceClass
   @memoized
   BuiltList<String> get genericBounds =>
       BuiltList<String>(element.typeParameters
-          .map((element) => (element.bound ?? '').toString()));
+          .map((element) => DartTypes.getName(element.bound) ?? ''));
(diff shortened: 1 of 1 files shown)


davidmorgan:
Thanks. Is `toString` going to change at some point? Otherwise, any suggestions for how I can catch regressions here?

scheglov:
I'd like to change `toString()` to return `DartType` presentation with nullability suffixes, e.g. `String* Function()` or `int? Function(String)`. So, it would be not what it used to return, and break clients that use it for code generation. We have never promised that it would suit this goal, but we have to fix anything that would break our internal repo.

My way to catch this will not help you - I just change `TypeImpl.toString()` to throw an exception.

davidmorgan:
Understood, thanks. I'll aim to pull this into google3 tomorrow--after that I'll be OOO for a week, how urgent is it that I succeed in landing in google3 tomorrow? I will adjust my effort accordingly ;) thanks.

scheglov:
Nothing is burning here, it is mostly for our convenience during analyzer development and debugging. It can wait for a week / 10 days.

PR #836, opened 2020-03-02
Make builder factory check `implName`-based

### LINKED ISSUE #835: Incorrect factory validation on _Builders
A custom builder `_Builder` will fail validation with `factory _Builder() = _$Builder`, asking instead for `_$_Builder`. However, it is the former that is generated.
```
Error in BuiltValueGenerator for

  abstract class _HomeSelectionSet with Home
    implements Built<_HomeSelectionSet, _HomeSelectionSetBuilder>.

Please make the following changes to use BuiltValue:
  1. Make builder class have exactly one factory:
    factory _HomeSelectionSetBuilder() = _$_HomeSelectionSetBuilder
```

(Not much of) a workaround is to use the `_$_Builder` statement for generation, then `s/_\$_/_$/g` in the source file, or simply don't use protected builders.

micimize:
closes #835 (I believe)

I just edited on github and haven't tested yet. Will test with a git dep shortly

### DIFF
--- a/built_value_generator/lib/src/serializer_source_class.dart
+++ b/built_value_generator/lib/src/serializer_source_class.dart
@@ -217,21 +217,38 @@ abstract class SerializerSourceClass
   CompilationUnitElement get compilationUnit =>
       element.library.definingCompilationUnit;
 
+  /// Returns the serializer class name for the generated implementation. If the manually
+  /// maintained class is private then we ignore the underscore here, to avoid
+  /// returning a class name starting `_$_`.
+  @memoized
+  String get serializerImplName {
+    final unhidden = name.startsWith('_') ? name.substring(1) : name;
+    return '_\$${unhidden}Serializer';
+  }
+
+  /// Returns the serializer instance name for the generated implementation,
+  /// referenced by the user as a `static serializer` attribute.
+  ///
+  /// If the manually maintained class is private then we ignore the underscore here, to avoid
+  /// returning a class name starting `_$_`.
+  @memoized
+  String get serializerInstanceName {
+    final unhidden = name.startsWith('_') ? name.substring(1) : name;
+    final camelCaseName =
+        unhidden.substring(0, 1).toLowerCase() + unhidden.substring(1);
+    return '_\$${camelCaseName}Serializer';
+  }
+
   Iterable<String> computeErrors() {
     var result = <String>[];
 
     if (!serializerSettings.custom) {
-      final camelCaseName =
-          name.substring(0, 1).toLowerCase() + name.substring(1);
-
       final expectedSerializerDeclaration =
-          'static Serializer<$genericName> get serializer => '
-          '_\$${camelCaseName}Serializer;';
+          'static Serializer<$genericName> get serializer => ${serializerInstanceName};';
       // We used to recommend raw types; recommend the full type now, but still
       // allow raw types.
       final expectedSerializerDeclarationRaw =
-          'static Serializer<$name> get serializer => '
-          '_\$${camelCaseName}Serializer;';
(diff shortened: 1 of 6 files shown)


micimize:
tested and checks/generators are corrected

micimize:
Just started implementing serialization and found what looks like a related issue there:
```
1. Declare _DimensionsSelectionSet.serializer as: static Serializer<_DimensionsSelectionSet> get serializer => _$_DimensionsSelectionSetSerializer; got static Serializer<_DimensionsSelectionSet> get serializer => _$dimensionsSelectionSetSerializer;
```

micimize:
Ok, applied the same changes to serializers!

micimize:
@davidmorgan anything else I need to do to move this forward?

davidmorgan:
Sorry, I'm pretty snowed under at the moment.

I'll make sure this lands before the next `built_value` release--but, not sure right now when that will be!

Thanks for your patience :)

built_value_generator/lib/src/serializer_source_class.dart
```diff
@@ -217,26 +217,45 @@ abstract class SerializerSourceClass
...
-          '_\$${camelCaseName}Serializer;';
+          'static Serializer<$name> get serializer => ${serializerInstanceName};';
       if (serializerDeclaration != expectedSerializerDeclaration &&
           serializerDeclaration != expectedSerializerDeclarationRaw) {
         result.add('Declare $name.serializer as: '
             '$expectedSerializerDeclaration got $serializerDeclaration');
       }
+      if (name.startsWith('_')) {
+        result.add('Cannot generate serializers for private class ${name}');
+      }
```
davidmorgan, line 258: Looks like you can revert most of the changes to this class, since with this check generation for private classes is no longer needed?

built_value_generator/test/built_value_generator_test.dart
```diff
@@ -745,6 +745,21 @@ abstract class Value implements Built<Value, ValueBuilder> {
         contains(
             '1. Set `nestedBuilders: false` in order to use `comparableBuilders: true`.'));
   });
+
+  test('Cleans generated class names for private classes', () async {
```
davidmorgan, line 749: Please also add a private class to `end_to_end_test/lib/values.dart`.

built_value_generator/lib/src/serializer_source_class.dart
micimize: I could, but I don't really see the advantage – except maybe clarifying the doc comments. IMO leaving the `_` handling makes it more consistent with the rest of the code base, but with an explicit decision to disallow something, rather than a check to prevent something unimplemented.

We _can_ do this but we _aren't_ vs we _could_ do this but we _haven't_.
Matter of taste though – we can revert the changes if you prefer.

built_value_generator/test/built_value_generator_test.dart
micimize: Done!

built_value_generator/lib/src/serializer_source_class.dart
davidmorgan: I think somewhere in the middle is best :)

Extracting `serializerImplName` and `serializerInstanceName` is an improvement.

But, the `name.startsWith` check on line 256 guarantees that the corresponding checks in those two methods will always return false in any real situation, making that codepath unused and untested. Untested code is better removed if it can be :)

Thanks!

micimize: @davidmorgan alright, cleaned up the dead code and rebased against master 👍

davidmorgan:
Looks good, thanks! I'll add a CHANGELOG entry. Please let me know if you'd like an entry in AUTHORS, or just send a PR.

PR #852, opened 2020-04-13
Enum const numbers

### LINKED ISSUE #657: Support numbers in @BuiltValueEnumConst
A small feature request.

In TypeScript it is extremely useful that you can give your enums either a string or number value. For example, we use Firestore, and when creating security rules, a number based value for figuring out what permissions are, is a given. In our old system this came from a `Role` enum, with numbers as values (Staff = 10, Admin = 20, Owner = 30. The gaps allow for expansion).

Example enum:
```
part 'role.g.dart';

class Role extends EnumClass {
  @BuiltValueEnumConst(wireName: 30)
  static const Role OWNER = _$owner;

  @BuiltValueEnumConst(wireName: 20)  
  static const Role ADMIN = _$admin;

  @BuiltValueEnumConst(wireName: 10)
  static const Role STAFF = _$staff;

  const Role._(String name) : super(name);

  static Serializer<Role> get serializer => _$roleSerializer;
  static BuiltSet<Role> get values => _$values;
  static Role valueOf(String name) => _$valueOf(name);
}
```

davidmorgan:
Fixes #657

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,9 @@
 # Changelog
 
-# 7.0.10 (unreleased)
+# 7.1.0 (unreleased)
 
+- Support serializing enums to ints: add `wireNumber` to
+  `@BuiltValueEnumConst`.
 - Trim `built_value_test` dependencies: depend on `matcher` instead of `test`.
 - Fix enum generator error messages when `value` and `valueOf` are missing.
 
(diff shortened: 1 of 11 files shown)


davidmorgan:
Thanks! Yes, you can mix both within a class. Added that to the end to end test so it's clear.

PR #889, opened 2020-07-03
Enable nnbd

davidmorgan:


built_value/lib/built_value.dart
```diff
@@ -309,25 +309,25 @@ int _indentingBuiltValueToStringHelperIndent = 0;
 
 /// A [BuiltValueToStringHelper] that produces single line output.
 class FlatBuiltValueToStringHelper implements BuiltValueToStringHelper {
-  StringBuffer _result = StringBuffer();
+  StringBuffer? _result = StringBuffer();
```
iinozemtsev, line 312: how it can be nullable?

davidmorgan: It's nullable because we set explicitly it to null when `build` is called; I can't quite figure out exactly why, though. That prevents you from calling `build` twice, but I have no idea why that's important.

Anyway, no change from the current implementation/API, which works for me :)

PR #893, opened 2020-07-14
Generate nnbd values 2

davidmorgan:
Add the rest of the "values" tests and generator changes.

Note, this PR is on top of: https://github.com/google/built_value.dart/pull/891


built_value/pubspec.yaml
```diff
@@ -6,7 +6,7 @@ description: >
 homepage: https://github.com/google/built_value.dart
 
 environment:
-  sdk: '>=2.9.0 <2.10.0'
+  sdk: '>=2.9.0-18.0 <2.10.0'
```
michalt, line 9: I'm guessing that will need to change soon ;)

built_value_generator/lib/src/serializer_source_class.dart
```diff
@@ -72,6 +73,18 @@ abstract class SerializerSourceClass
   @memoized
   String get name => element.name;
 
+  // TODO(davidmorgan): this is not the right way to check, but is good enough
```
michalt, line 76: Especially since 2.9 will not work much longer ;)

end_to_end_test/lib/enums_nnbd.dart
```diff
@@ -0,0 +1,104 @@
+// Copyright (c) 2017, Google Inc. Please see the AUTHORS file for details.
+// All rights reserved. Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+// @dart=2.9
```
michalt, line 4: Out of curiosity -- is this necessary? Wouldn't this get Null Safety by default and you only need to mark the legacy parts as `2.8`?

built_value_generator/lib/src/serializer_source_class.dart
```diff
@@ -72,6 +73,18 @@ abstract class SerializerSourceClass
...
+  // while developing. Fix before publishing.
+  @memoized
+  bool get isNonNullByDefault =>
+      element.source.contents.data.contains('// @dart=2.9');
+
+  @memoized
+  String get orNull => isNonNullByDefault ? '?' : '';
+
+  @memoized
+  String get notNull => isNonNullByDefault ? '!' : '';
```
michalt, line 86: Is `notNull` used? (I don't see it...)

built_value/pubspec.yaml
davidmorgan: Likely :)

Doesn't have any lasting effect until publishing, though, and I won't publish this until there is a fixed version with null safety support. So, nothing to worry about for now :)

end_to_end_test/lib/enums_nnbd.dart
davidmorgan: Yes, although the 2.9 means that if it's not using the new version for some reason we get an error, which is handy.

Neither option seems perfect though :)

built_value_generator/lib/src/serializer_source_class.dart
davidmorgan: Not yet, I likely copied it in from a later PR :)

davidmorgan:
Thanks! Next one headed your way soon...

PR #895, opened 2020-07-23
Generate nnbd interfaces

davidmorgan:


built_value_generator/lib/src/serializer_source_class.dart
```diff
@@ -506,7 +507,7 @@ class $serializerImplName implements PrimitiveSerializer<$genericName> {
           compilationUnit, genericParameters.toBuiltSet());
       final cast = field.generateCast(compilationUnit, _genericBoundsAsMap);
       if (field.builderFieldUsesNestedBuilder) {
-        if (builtValueSettings.autoCreateNestedBuilders && !hasBuilder) {
+        if (builtValueSettings.autoCreateNestedBuilders || hasBuilder) {
```
michalt, line 510: Could you help me understand this change? (I'm neither entirely sure what are its consequences nor exactly why it's necessary ;) Thanks!

davidmorgan: Yeeeah this is an unpleasant, almost-not-used corner case :)

`hasBuilder` means the user has opted to write the builder outline explicitly rather than have the builder be fully generated.

And this condition is about whether the nested builder might be null. If it might be null, we can't call `replace` on it, we have to make sure it's not null, which is what the next `else` case does.

By default, with `autoCreateNestedBuilders`, the nested builder can't be null, so it's safe to call `replace`.

Previously, I was assuming that if you write your own builder then you might allow the nested builder to be nullable, and so `&& !hasBuilder` to fall through to the 'might be null' case. But, it turns out that this doesn't work for collections, the serializer doesn't give them the right generic type and so the cast fails.

Fixing this is hard, so I revert to the old behaviour--which actually is the behaviour in the latest published version--which is that you have to make the builder in the explicit builder class not be nullable, and we always call `replace`.

tl;dr: this was sort of broken before the null safety changes, and now it's as it was before ;) which is not great, but a correct fix is hard, and AFAICT nobody in google3 is using this case anyway :D

And this is why almost-unused corner cases in codegen are not fun ;)

michalt: Cool, thanks for the explanation! :)

PR #908, opened 2020-08-20
Remove unused dart:async import

MichaelRFairhurst:
Since this package does not support Dart 2.0, this import is not
necessary.


davidmorgan:
Thanks :)

PR #931, opened 2020-10-23
Convert List<dynamic> to List<Object> in standard_json_plugin

iinozemtsev:
Otherwise json deserialization using standard_json_plugin fails in null safe mode

### DIFF
--- a/built_value/lib/standard_json_plugin.dart
+++ b/built_value/lib/standard_json_plugin.dart
@@ -63,6 +63,11 @@ class StandardJsonPlugin implements SerializerPlugin {
       } else {
         return _toList(object, _needsEncodedKeys(specifiedType));
       }
+    } else if (object.runtimeType == _dynamicListRuntimeType) {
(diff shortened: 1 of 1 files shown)


built_value/lib/standard_json_plugin.dart
```diff
@@ -50,6 +50,11 @@ class StandardJsonPlugin implements SerializerPlugin {
       } else {
         return _toMap(object, _needsEncodedKeys(specifiedType));
       }
+    } else if(object is List<dynamic>) {
```
davidmorgan, line 53: This matches too much; 'is List<dynamic>' returns true for every list.

I think instead you'll want to check if object.runtimeType == <dynamic>[].runtimeType. It probably makes sense to cache <dynamic>[].runtimeType in a static field.

iinozemtsev: Done. I've also noticed that I have patched a wrong place (afterSerialize instead of beforeSerialize), fixed

PR #954, opened 2020-12-11
Fix: escape serializer code

kuhnroyal:
This fixes usages like `@BuiltValueEnumConst(wireName: r'$foo')` which currently generates `'$foo'` in `_toWire/_fromWire`.
With this change it correctly generates `'\$foo'`.

I wonder if instead of escaping all of these could be raw strings?

### DIFF
--- a/built_value_generator/lib/src/serializer_source_class.dart
+++ b/built_value_generator/lib/src/serializer_source_class.dart
@@ -419,7 +419,7 @@ class $serializerImplName implements PrimitiveSerializer<$genericName> {
 
   static String _toCode(Object object) {
     if (object is String) {
-      return "'$object'";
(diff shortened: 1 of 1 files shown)


davidmorgan:
Sorry for slow response--thanks. Please add an end to end test.

Unfortunately it's not currently possible to publish `built_value_generator` because of the move to null safety and a circular dependency with `package:build`; hopefully I can resolve this in January.

davidmorgan:
Fixed with a test in https://github.com/google/built_value.dart/pull/996

kuhnroyal:
Thanks, I didn't get around to it yet, sorry!

davidmorgan:
No worries, thanks for the fix :)

PR #975, opened 2021-01-28
Remove workaround for test folders.

davidmorgan:


emmanuel-p (approved):

davidmarne:
@davidmorgan could you please release the 8.x line again when possible to include this change? I have some models in a test directory, that are used only for tests, that do not get generated with nnbd when using `8.0.0-nullsafety.0`

PR #1015, opened 2021-03-25
Fix `Null check operator used on a null value`

hoc081098:
Fixes #1013

PR #1027, opened 2021-04-27
feat: re-added support for serializeNulls

### LINKED ISSUE #604: null values are stripped when serializing

larssn:
This basically uses most of the old code, with a few exceptions.

I have no idea how to test this code without modifying the .pub-cache, since most of the packages references external dependencies. So it probably shouldn't be merged before someone explains how to do that.

Closes #604


dave26199:
Try tool/set_local_version then tool/presubmit ... sorry for brief reply,
posting from phone ... will try to take a look this week.

larssn:
Thanks for the tip!

It works, however it's close to how it was before. Might need a more thorough solution for NNBD.

larssn:
Don't merge, there seems to be a problem serializing lists that can be null.

larssn:
Added end to end tests for `null` lists and maps (which fail).

Doesn't seem trivial to fix those two. And I fear the only workaround for now is not use `serializeNulls: true` and lists/maps together.

larssn:
Testing the code in our app reveals even more problems. Gonna close this as its not viable.

larssn:
@davidmorgan We've been testing this PR together with #1015, and we've not had any problems.

PR #1047, opened 2021-06-08
Fix bug with nested nullable generics

### LINKED ISSUE #1045: Nested nullable generics
Impossible to build fields with nested nullable generics. Example: 

```dart
BuiltList<BuiltList<String>?> get hobbies;
BuiltList<BuiltList<String>?>? get hobbies;
```

This will throw a `FormatterException`:
```
[SEVERE] built_value_generator:built_value on lib/src/aaa.dart:
An error `FormatterException` occurred while formatting the generated source for
  `package:test_generator_fix/src/aaa.dart`
which was output to
  `lib/src/aaa.built_value.g.part`.
This may indicate an issue in the generator, the input source code, or in the
source formatter.
Could not format because the source could not be parsed:

line 49, column 117 of .: Expected an identifier.
   ╷
49 │       value, specifiedType: const FullType(BuiltList, const [const FullType(BuiltList, const [const FullType(String>)])]))! as BuiltList<Object>);
   │                                                                                                                     ^
   ╵
line 17, column 426 of .: Expected an identifier.
```

mpkander:
Fix #1045

### DIFF
--- a/built_value_generator/lib/src/dart_types.dart
+++ b/built_value_generator/lib/src/dart_types.dart
@@ -69,7 +69,8 @@ class DartTypes {
         return dartType.element.name + suffix;
       } else {
         final typeArgumentsStr = typeArguments
-            .map((type) => getName(type, withNullabilitySuffix: true))
+            .map((type) =>
+                getName(type, withNullabilitySuffix: withNullabilitySuffix))
             .join(', ');
         return '${dartType.element.name}<$typeArgumentsStr>$suffix';
       }


davidmorgan:
Thanks for finding the fix. It'll need test coverage :) feel free to take a shot at adding some, or I can take it over. Thanks.

davidmorgan:
Digging into this, it doesn't look like the right fix--I'll see if I can come up with the right fix :) thanks.

PR #1072, opened 2021-06-17
Error on SerializersFor wrong type.

### LINKED ISSUE #635: Serializers class is only generated if Serializers type is specified
```dart
// Does not build the serializers class and doesn't provide any warning.
@SerializersFor(const [
  Foo,
])
final _serializers = _$_serializers;

// Does build the serializer class.
@SerializersFor(const [
  Foo,
])
final Serializers _serializers = _$_serializers;
```

It took me awhile to figure out why it wasn't working. If the code generator is going to be that picky, it would be nice if it at least gave me an error on the first example there, which it could know to do given the annotation.

davidmorgan:
Fixes #635

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@
 - Fix enum deserialization fallback for `int`.
 - Removed Angular mixin from example, as this feature is no longer needed:
   Angular now directly supports using static members in templates.
+- Annotating a wrong type getter with `@SerializersFor` is now an error,
+  instead of just generating nothing.
 - Add `@BuiltValueHook` annotation. It provides the same functionality as
   `_initializeBuilder` and `_finalizeBuilder`, but in a more visible way:
   annotate a static method on the value class with `@BuiltValueHook` to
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -17,6 +17,8 @@
   annotation is used where it has no effect.
 - Removed Angular mixin from example, as this feature is no longer needed:
   Angular now directly supports using static members in templates.
+- Annotating a wrong type getter with `@SerialiersFor` is now an error,
```
iinozemtsev, line 20: typo `Serialiers`

built_value_generator/lib/src/serializer_source_library.dart
```diff
@@ -133,11 +158,27 @@ abstract class SerializerSourceLibrary
...
   /// Generates serializer source for this library.
   String generateCode() {
-    var errors = concat(sourceClasses
-        .map((sourceClass) => sourceClass.computeErrors())
-        .toList());
+    var errors = [
+      ...computeErrors(),
+      ...concat(sourceClasses
+          .map((sourceClass) => sourceClass.computeErrors())
+          .toList())
```
iinozemtsev, line 180: you can do expand instead of concat(map), or with fancy literals something like this:

```
for (var sourceClass in sourceClasses)
   ...sourceClass.computeErrors()
```

iinozemtsev (approved):

CHANGELOG.md
davidmorgan: Done.

built_value_generator/lib/src/serializer_source_library.dart
davidmorgan: Done, thanks!

PR #1074, opened 2021-06-18
Support polymorphism with mixed in parent value type.

### LINKED ISSUE #1071: Support Built mixins
I'd like to be able to do this:

```
@BuiltValue(nestedBuilders: false, instantiable: false)
abstract class Parent {
  BuiltList<Object> get noNestedBuilder;
}

@BuiltValue(nestedBuilders: false)
abstract class Child with Parent implements Built<Child, ChildBuilder> {
  ...
}
```

It requires:
* respect `nestedBuilders` in generated abstract builder classes
* pick up mixins similarly to implemented interfaces and make generated builders mix in the parent builders again

There is a workaround for the second issue above by saying `abstract class Child with Parent implements Built<Child, ChildBuilder>, Parent`. It will do `ChildBuilder implements ParentBuilder` rather than `ChildBuilder with ParentBuilder`, but that's OK if there are only generated builders at play.

Google specific: The necessary fixes are outlined in cl/379939823.

davidmorgan:
Fixes #1071

### DIFF
--- a/end_to_end_test/lib/polymorphism_nnbd.dart
+++ b/end_to_end_test/lib/polymorphism_nnbd.dart
@@ -232,6 +232,18 @@ abstract class Car implements Vehicle, Built<Car, CarBuilder> {
   int get seatsCount;
 }
 
+// Check that generated builders mix in parent buildes when the class mixes
+// in the other class; otherwise, they won't build.
+abstract class MixinCar
+    with Vehicle
+    implements Built<MixinCar, MixinCarBuilder> {
+  MixinCar._();
+
+  factory MixinCar([void Function(MixinCarBuilder)? updates]) = _$MixinCar;
+
+  int get seatsCount;
+}
+
 abstract class VehicleColor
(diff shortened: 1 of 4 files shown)


end_to_end_test/lib/polymorphism_nnbd.dart
```diff
@@ -232,6 +232,18 @@ abstract class Car implements Vehicle, Built<Car, CarBuilder> {
   int get seatsCount;
 }
 
+// Check that generated builders mix in parent buildes when the class mixes
+// in the other classs; otherwise, they won't build.
```
iinozemtsev, line 236: typo `classs`

iinozemtsev (approved):

end_to_end_test/lib/polymorphism_nnbd.dart
davidmorgan: Thanks :)

PR #1085, opened 2021-08-03
Add support for analyzer 2.0.0

kuhnroyal:
Analyzer 2.0.0 seems to work, tests succeed.
Would be good to get this released.

### DIFF
--- a/built_value_generator/lib/built_value_generator.dart
+++ b/built_value_generator/lib/built_value_generator.dart
@@ -62,7 +62,7 @@ class BuiltValueGenerator extends Generator {
           st);
     }
 
-    for (var element in libraryElement.units.expand((unit) => unit.types)) {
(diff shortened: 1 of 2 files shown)


simolus3:
I think you can fix the analysis warnings from the CI by replacing this line:

https://github.com/google/built_value.dart/blob/7cec575a2614f7a25bd74ca6d9d95cf5cff7d93e/built_value_generator/lib/built_value_generator.dart#L65

with

```dart
 for (var element in libraryElement.units.expand((unit) => unit.classes)) { 
```

The analyzer constraint should then be changed to `>=1.8.0 <3.0.0` since `classes` has been added in 1.8.0.

kuhnroyal:
Thanks @simolus3, completely forgot to follow up on this PR.

davidmorgan:
Thanks!

Issue #941, opened 2020-11-02
Issue with analyzer 0.40.5, InconsistentAnalysisException


alanknight-wk:
I haven't dug into this more deeply, but in attempting to run with bleeding edge 2.12 and analyzer 0.40.5 I encountered the following error building one of our packages. It may be something that's going to be an issue in upcoming builds.

[SEVERE] built_value_generator:built_value on lib/src/data/models/attachment.sg.dart:
Error in BuiltValueGenerator for abstract class Attachment implements Built<Attachment*, dynamic>*.
InconsistentAnalysisException: Requested result might be inconsistent with previously returned results

davidmorgan:
Thanks Alan. That's a new one to me ... @scheglov could you give any pointers on why we might get `InconsistentAnalysisException`, please?

kuhnroyal:
See https://github.com/google/built_value.dart/issues/923#issuecomment-721138364

Also I now managed to generate with `analyzer: 0.40.5` the problem was caused by one of `build > 1.3.0, build_runner > 1.10.2 or source_gen > 0.9.5` but I can't nail it down exactly, might be a combination of those that fails.

kuhnroyal:
Yea, as soon as I set `build_runner: 1.10.3` it fails, this also pulls in `build: 1.5.0`.

scheglov:
You would get `InconsistentAnalysisException` if you try to call anything on a `AnalysisSession` that was asked from `AnalysisDriver` at time `T`, if you called `AnalysisDriver.changeFile(path)` or `AnalysisDriver.addFile(path)` after `T`.

davidmorgan:
@scheglov I'm stuck--any chance you could take a look please?

Repro here: https://github.com/davidmorgan/built_value.dart/tree/publishable-null-safety
Then in folder `end_to_end_test` run `pub run build_runner build` to get

```
Precompiling executable... (14.2s)
Precompiled build_runner:build_runner.
[INFO] Generating build script completed, took 260ms
[INFO] Reading cached asset graph completed, took 61ms
[INFO] Checking for updates since last build completed, took 437ms
[INFO] Running build completed, took 17ms
[INFO] Caching finalized dependency graph completed, took 43ms
[SEVERE] built_value_generator:built_value on lib/mixins_nnbd.dart (cached):
[...]

PR #1090, opened 2021-08-25
Back-patch v7 to widen analyzer range and workaround InconsistentAnalysisException

evanweible-wf:
(Moved from https://github.com/evanweible-wf/built_value.dart/pull/1)

_Note: this PR is currently opened against master, but the branch is based off of the `v7.1.0` tag._

The built_value_generator package at version 7.1.0 depends on `analyzer: '>=0.39.3 <0.40.0'`, which prevents consumers from resolving newer versions of certain build packages. However, built_value_generator is completely compatible with analyzer v0.40.x as-is. So, to allow consumption of built_value* 7.x with newer build packages without also having to upgrade to Dart 2.12+, this commit widens the analyzer range to include v0.40.x.

I've verified locally that on both Dart 2.7.2 and 2.13.4:
- `pub get` resolves
- `./tool/presubmit` builds all packages successfully
- `pub run test` passes on `built_value_generator`

I also included a commit to cherry-pick the workaround for #941 that was added to v8, since we still hit that in a few places where we can't pin `build_runner` to the right version.

---

Some context: Workiva is about to upgrade from Dart 2.7 to 2.13 (for a few reasons, we were unable to upgrade to versions prior to 2.12). So far, we've been able to upgrade all of our individual packages to run on 2.13 successfully while still supporting dependency ranges that resolve on both 2.7 and 2.13. This buys us the ability to build our main application on both versions and switch which version is deployed with a flag.

One of the critical dependency versions that we need to resolve to on 2.13 is `build_web_compilers v2.12.0`, because prior to this version, code is incorrectly opted-in to NNBD even though the package's minimum Dart SDK is less than 2.12. Unfortunately, `built_value_generator v7.1.0` prevents us from resolving to that version because of its analyzer dependency. Because of that, we had been using a `>=7.1.0 <9.0.0` range which we've now learned is unsafe because code generated with built_value_generator v8 (reasonably) references code that does not exist in built_value v7.

If this is an acceptable back-patch, we would be able to depend on built_value and built_value_generator `^7.1.1` _and_ resolve `build_web_compilers 2.12.0` on Dart 2.13, which means we'd be able to complete this upgrade smoothly.

@davidmorgan

### DIFF
--- a/built_value/lib/built_value.dart
+++ b/built_value/lib/built_value.dart
@@ -327,7 +327,7 @@ class FlatBuiltValueToStringHelper implements BuiltValueToStringHelper {
 
   @override
   String toString() {
-    _result..write('}');
+    _result.write('}');
     var stringResult = _result.toString();
     _result = null;
     return stringResult;


davidmorgan:
Thanks Kevin :)

@evanweible-wf do you need me to publish this to pub, or are you already good with having it on a branch?

evanweible-wf:
@davidmorgan Kevin published it to pub already, so we're good to go!

davidmorgan:
@evanweible-wf ahh, I'm so used to `built_value` and `built_value_generator` being published together that I only checked `built_value`, I see v7.1.1 for `built_value_generator`. LG. Thanks!

PR #1096, opened 2021-10-14
Makes `rebuild` take a void function.

lrhn:
It's not very clear whether a value returned by the `updates` parameter to `rebuild` means anything.
All examples use `=>` notation which returns the builder. The return type is currently `dynamic`, which *suggests* that the value is used for something, potentially something which cannot even be typed.

Marking the parameter as a `void` returning function makes it clear that you should not worry about the return value.

### DIFF
--- a/built_value/lib/built_value.dart
+++ b/built_value/lib/built_value.dart
@@ -18,7 +18,7 @@ abstract class Built<V extends Built<V, B>, B extends Builder<V, B>> {
   ///
   /// The implementation of this method will be generated for you by the
   /// built_value generator.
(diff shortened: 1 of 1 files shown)


davidmorgan:
Thanks for pinging Will.

I'd have to release as a major version increase, i.e. `built_value 9`, I'm not sure that's worth it just for this change; but if I make a breaking release for other reasons I'll include this one.

Note to self: we actually cleaned up google3 for violations and apply this change as a patch; there were a nontrivial number of violations, so I expect there are violations in the wild, too; they were of the form `(b) { ...; return ...; }`

PR #1107, opened 2022-01-18
Make generated deserialize method not throw cast_nullable_to_non_nullable lint warning

### LINKED ISSUE #1101: Generated code produces lint warning: cast_nullable_to_non_nullable
This is low priority, because I think most people exclude generated files from analysis. I wouldn't even see this if my exclude wasn't incorrectly setup just now.

The lint warning [is listed](https://dart-lang.github.io/linter/lints/cast_nullable_to_non_nullable.html) as experimental (as of December 16, 2021) but seems to be included by default in `package:lint`. It does seem like a good practice:

| Don't cast a nullable value to a non nullable type 

Screen Shot 2021-12-16 at 10 20 33

Right now, built_value generates deserialization code like this:

```dart
  @override
  Something deserialize(
      Serializers serializers, Iterable<Object?> serialized,
      {FullType specifiedType = FullType.unspecified}) {
    final result = new SomethingBuilder();

    final iterator = serialized.iterator;
    while (iterator.moveNext()) {
      final key = iterator.current as String;   // LINT HERE
      iterator.moveNext();
      final Object? value = iterator.current;
      switch (key) {
        case 'blah':
          result.blah = serializers.deserialize(value,
              specifiedType: const FullType(bool)) as bool;   // LINT HERE
          break;
      }
    }

    return result.build();
  }
```

The fix would be to add a null assert, for example:

```dart
final key = iterator.current! as String;
```

HelioStrike:
Fixes #1101.

The change adds a couple' null asserts to `deserialize` to prevent cast_nullable_to_non_nullable lint warning.

### DIFF
--- a/built_value_generator/test/serializer_generator_test.dart
+++ b/built_value_generator/test/serializer_generator_test.dart
@@ -226,6 +226,27 @@ abstract class _Value implements Built<_Value, _ValueBuilder> {
 }
 '''), contains(r'1. Cannot generate serializers for private class _Value'));
     });
+
+    test('deserialize is null safe', () async {
+      final generated = await generate(r'''
+        library value;
+
+        import 'package:test_support/test_support.dart';
+
+        part 'value.g.dart';
+
+        abstract class Value implements Built<Value, ValueBuilder> {
+          static Serializer<Value> get serializer => _$valueSerializer;
+          bool get aBool;
+          
+          Value._();
+          factory Value([void Function(ValueBuilder) updates]) = _$Value;
+        }
+        ''');
+      expect(generated, contains(r'final key = iterator.current! as String;'));
+      expect(generated,
+          contains(r'specifiedType: const FullType(bool))! as bool;'));
+    });
   });
 }
 
--- a/built_value_generator/test/serializer_generator_test.dart
+++ b/built_value_generator/test/serializer_generator_test.dart
@@ -226,27 +226,6 @@ abstract class _Value implements Built<_Value, _ValueBuilder> {
 }
 '''), contains(r'1. Cannot generate serializers for private class _Value'));
     });
-
-    test('deserialize is null safe', () async {
-      final generated = await generate(r'''
-        library value;
-
-        import 'package:test_support/test_support.dart';
-
-        part 'value.g.dart';
-
-        abstract class Value implements Built<Value, ValueBuilder> {
-          static Serializer<Value> get serializer => _$valueSerializer;
-          bool get aBool;
-          
-          Value._();
-          factory Value([void Function(ValueBuilder) updates]) = _$Value;
-        }
-        ''');
-      expect(generated, contains(r'final key = iterator.current! as String;'));
-      expect(generated,
-          contains(r'specifiedType: const FullType(bool))! as bool;'));
-    });
   });
 }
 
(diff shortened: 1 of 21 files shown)


HelioStrike:
Should I re-generate the generated files with the changed code?

built_value_generator/test/serializer_generator_test.dart
```diff
@@ -226,6 +226,27 @@ abstract class _Value implements Built<_Value, _ValueBuilder> {
 }
 '''), contains(r'1. Cannot generate serializers for private class _Value'));
     });
+
+    test('deserialize is null safe', () async {
```
davidmorgan, line 230: Thanks for the PR :)

Please add this as an end to end test rather than a unit test; the setup is that correct usage is tested as end to end tests, so we test the actual effect of the feature, and wrong usage is tested as unit tests, so we test the error output.

To add as an end to end test:

Run tool/use_local_deps so the end to end test uses the updated version of the generator. In the end_to_end_test folder you can now run the generation to iterate.

Add the lint to analysis_options.yaml for end_to_end_test. The lints are checked on presubmit, so this is now a test for the feature. Probably you don't need to add any new actual end to end cases to cover the feature.

Once it looks like it's working, try tool/presubmit. That regenerates everything, checks that it passes analysis (including lints), and runs all the tests. It's a bit slow though :)

HelioStrike: I had to change a couple' small things and got the tool/presubmit to pass. That's a pretty neat test setup :)

Should I update the PR with newly generated code (after tool/presubmit) and the analysis_options change too, or just the little change I had to make to make tool/presubmit pass?

dave26199: Please include everything in the pr :) thanks.

HelioStrike: Updated the PR. Thanks!

HelioStrike: Also, the presubmit tool seems to be using dartfmt, pub, and dartanalyzer which seem to be replaced with 'dart format', 'dart pub', and 'dart analyze' for the outside world. I think the recent push also contains these changes. Let me know if that's not ideal and I'll change the presuibmit tool script back.

davidmorgan: Works for me, thanks :)

Change looks good, please also add to the changelog, something like:

# 8.2.0 (unreleased)

- Make generator output additional explicit null checks so the generated code complies with the `cast_nullable_to_non_nullable` lint.

HelioStrike: Copy pasta and.... updated! Thanks! :)

davidmorgan: Ah, sorry, one more, please remove the unit test :) ... the e2e test has this covered now, and unit tests on generated code are quite painful to maintain :) thanks!

HelioStrike: Done, and no worries, thanks for being patient and pointing out all these little changes. :)

davidmorgan: LG, thanks :)

Issue #1093, opened 2021-09-15
final params in Enum


aarajput:
I am getting below error when trying to use final parameters in enum and trying to generate code

Please make the following changes to use EnumClass:
1. Have exactly one constructor: const Environments._(String name) : super(name);
2. Add method: static Environments valueOf(String name) => _$valueOf(name)

```
import 'package:built_collection/built_collection.dart';
import 'package:built_value/built_value.dart';
import 'package:built_value/serializer.dart';

part 'environments.g.dart';

class Environments extends EnumClass {
  static Serializer<Environments> get serializer => _$environmentsSerializer;

  static BuiltSet<Environments> get values => _$values;

  static Environments valueOf(final String name) => _$valueOf(name);

  static const Environments localhost = _$localhost;
  static const Environments development = _$development;
  @BuiltValueEnumConst(fallback: true)
  static const Environments production = _$production;

  const Environments._(final String name) : super(name);
}

```

davidmorgan:
Enums can't have fields--you'll need to instead add a getter that does a `switch` over `this`, i.e. have a method where you check the enum type and return the right value.

aarajput:
@davidmorgan ! I am not talking about using fields in enum, I am talking about using final parameters. Like this one
**static Environments valueOf(final String name) => _$valueOf(name);**

davidmorgan:
Ah, got it--thanks.

Indeed, that is not allowed/supported :) but you don't need it, either: adding final on a parameter makes no difference to the API, it is a hint to the implementation.

Is it possible you want it to meet a lint or some style guide?

aarajput:
In latest version of lint, they have added option of _prefer_final_parameters_ . I am using it. But in enum, i get error if i use final parameters.

davidmorgan:
That makes sense.

You can use `// ignore_for_file: prefer_final_parameters` for now--I will make this a feature request to allow `final`.

PR #1116, opened 2022-03-19
Support final parameters in EnumClass constructor and valueOf method.

HelioStrike:
Solves #1093.

This change allows using final params in a class inheriting from `EnumClass`, for its constructor, and `valueOf` method.

I also had to replace the usages of `interfaces2` with `interfaces` (due to deprecation) in [value_source_class.dart](https://github.com/google/built_value.dart/blob/master/built_value_generator/lib/src/value_source_class.dart) for tool/presubmit to pass.

### DIFF
--- a/built_value_generator/test/enum_class_generator_test.dart
+++ b/built_value_generator/test/enum_class_generator_test.dart
@@ -222,7 +222,7 @@ class TestEnum extends EnumClass {
 }
 '''), endsWith(r'''Please make the following changes to use EnumClass:
 
-1. Have exactly one constructor: const TestEnum._((final )?String name) : super(name);'''));
+1. Have exactly one constructor: const TestEnum._(String name) : super(name);'''));
       });
 
       test('with error on incorrect constructor', () async {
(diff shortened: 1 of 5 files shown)


built_value_generator/test/enum_class_generator_test.dart
```diff
@@ -222,7 +222,7 @@ class TestEnum extends EnumClass {
 }
 '''), endsWith(r'''Please make the following changes to use EnumClass:
 
-1. Have exactly one constructor: const TestEnum._(String name) : super(name);'''));
+1. Have exactly one constructor: const TestEnum._((final )?String name) : super(name);'''));
```
davidmorgan, line 225: Could you please tweak to keep the original failure message?

If people want "final" they will presumably try it without being asked :) but otherwise I think it's less confusing to highlight just the one way forward.

davidmorgan (commented):
Looks great, thanks! Just one suggestion.

built_value_generator/test/enum_class_generator_test.dart
HelioStrike: Makees sense, updated :)

davidmorgan (approved):
Thanks!

PR #1150, opened 2022-05-20
Implement support for optional generic bounds

MaxHaertwig:
Currently, the library doesn't support optional generic bounds. The optionality of the bounds is simply dropped, resulting in a built value that behaves differently than its abstract class.

The following code

```dart
abstract class Box<T extends Content?> implements Built<Box<T>, BoxBuilder<T>> {
   ...
}
```

results in the generation of

```dart
class _$Box<T extends Content> extends BoxValue<T> {
   ...
}

class BoxBuilder<T extends Content> implements Builder<Box<T>, BoxBuilder<T>> {
    ....
}
```

The correct behavior would be the retention of the question mark (`T extends Content?`). This pull request resolves this issue.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -10,3 +10,4 @@ John Dengis <jadengis@gmail.com>
 Resul Alkan/MetGlobal <me@resulalkan.com>
 Dao Hoang Son <dao@hoangson.vn>
 Ahmed Fwela/Bdaya Development <ahmednfwela@bdaya-dev.com>
+Max Härtwig/Google Inc. <maxhaertwig@google.com>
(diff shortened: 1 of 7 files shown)


davidmorgan:
Heh. `null is T ? value as T : ...` is a very entertaining code snippet; when I first saw it I was almost sure it could not be correct, but, I think is ;)

Looks great, thanks! Could you please 

 - rebase; I'll take care of the failing presubmit when I release
 - add a note to the changelog; doesn't matter what exactly, I'll reformat it when I release, just so I don't forget
 - add yourself to AUTHORS if you like

Thanks :)

MaxHaertwig:
Yeah, took me a while to figure this line out ;)

Done, thanks!

PR #1162, opened 2022-06-10
Fix erroneously generated null check for fields with generic bounds

MaxHaertwig:
The current implementation generates a null check for all fields with generic bounds, even if they have been declared as nullable. This fix resolved this issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Changelog
 
+- Fix erroneously generated null check for fields with generic bounds.
+
 # 8.3.2
 
 - Migrate `built_value_test` to null safety.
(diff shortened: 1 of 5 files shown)


davidmorgan:
Hmm, wit https://github.com/google/built_value.dart/pull/1163 so the end to end test uses the updated generator, I get

```
error • lib/generics_nnbd.g.dart:985:73 • The argument type 'T?' can't be assigned to the parameter type 'T'. • argument_type_not_assignable"
```

looking at the change to the generated code, from

```
  _$BoundNullableGenericValue<T> _build() {
    final _$result = _$v ??
        new _$BoundNullableGenericValue<T>._(
            value: null is T
                ? value as T
                : BuiltValueNullFieldError.checkNotNull(
                    value, r'BoundNullableGenericValue', 'value'));
    replace(_$result);
    return _$result;
  }
```

to

```
  _$BoundNullableGenericValue<T> _build() {
    final _$result = _$v ?? new _$BoundNullableGenericValue<T>._(value: value);
    replace(_$result);
    return _$result;
  }
```

I think it needs to keep the cast for this case?

PR #1174, opened 2022-07-11
Fix nested builder bug in custom builders, and allow define abstract getters/setters in custom builders

### LINKED ISSUE #1173: Generator breaks when a custom Builder is defined for a BuiltValue that contains a nested builder for a nullable BuiltValue
If you want to create a custom Builder and use a nested nullable BuiltValue in your object, it's currently impossible if you'd like to use nested builders.

Take this rudimentary Built class for example:
```
abstract class Value implements Built<Value, ValueBuilder> {
  // Note: This behavior can be reproduced with any nested BuiltValue. I'm only
  // using Value here since it's the simplest way to reproduce this bug.
  Value? get value;

  Value._();

  factory Value([void Function(ValueBuilder) updates]) = _$Value;
}
```

No scenario works where you can implement a custom builder with a nested builder and have it generate a valid output, despite auto-generated builders handling this scenario fine.

The only thing that works is to make the Builder class member for `value` be the Built type, which makes usage of the builder cumbersome:
```
abstract class ValueBuilder implements Builder<Value, ValueBuilder> {
  Value? value;

  ValueBuilder._();
  factory ValueBuilder() = _$ValueBuilder;
}
```

### Attempting to use a getter/setter pair (which would match existing, pre-null-safe behavior)
```
abstract class ValueBuilder implements Builder<Value, ValueBuilder> {
  ValueBuilder get value;
  set value(ValueBuilder? newValue);

  ValueBuilder._();
  factory ValueBuilder() = _$ValueBuilder;
}
```

```
49: return super.value;                                           The getter 'value' is always abstract in the supertype.
                                                                  #abstract_super_member_reference

53: set value(ValueBuilder value) {                               '_$ValueBuilder.value=' ('void Function(ValueBuilder)') isn't a valid
                                                                  override of 'ValueBuilder.value=' ('void Function(ValueBuilder?)').
                                                                  #invalid_override

55: super.value = value;                                          The setter 'value' is always abstract in the supertype.
                                                                  #abstract_super_member_reference

63: super.value = $v.value?.toBuilder();                          The setter 'value' is always abstract in the supertype.
                                                                  #abstract_super_member_reference

86: _$result = _$v ?? new _$Value._(value: super.value?.build()); The getter 'value' is always abstract in the supertype. (warning) The
                                                                  receiver can't be null, so the null-aware operator '?.' is unnecessary.
                                                                  #abstract_super_member_reference #invalid_null_aware_operator

91: super.value?.build();                                         The getter 'value' is always abstract in the supertype. (warning) The
                                                                  receiver can't be null, so the null-aware operator '?.' is unnecessary.
                                                                  #abstract_super_member_reference #invalid_null_aware_operator
```

### Attempting to use `late` (didn't expect this to work in the first place)
```
abstract class ValueBuilder implements Builder<Value, ValueBuilder> {
  late ValueBuilder value;

  ValueBuilder._();
  factory ValueBuilder() = _$ValueBuilder;
}
```

```
63: super.value = $v.value?.toBuilder();                          A value of type 'ValueBuilder?' can't be assigned to a variable of type
                                                                  'ValueBuilder'. #invalid_assignment

86: _$result = _$v ?? new _$Value._(value: super.value?.build()); (warning) The receiver can't be null, so the null-aware operator '?.' is
                                                                  unnecessary. #invalid_null_aware_operator

91: super.value?.build();                                         (warning) The receiver can't be null, so the null-aware operator '?.' is
                                                                  unnecessary. #invalid_null_aware_operator
```

### Attempting to make the nested builder nullable
```
abstract class ValueBuilder implements Builder<Value, ValueBuilder> {
  ValueBuilder? value;

  ValueBuilder._();
  factory ValueBuilder() = _$ValueBuilder;
}
```

```
Please make the following changes to use BuiltValue:

1. Make builder field value have type: Value? (or, if applicable, builder)
package:built_value_generator/src/value_source_class.dart 725  ValueSourceClass.generateCode
package:built_value_generator/built_value_generator.dart 67    BuiltValueGenerator.generate
===== asynchronous gap ===========================
package:source_gen/src/builder.dart 352                        _generate
```

### Attempting to explicitly define the getter and setter
```
abstract class ValueBuilder implements Builder<Value, ValueBuilder> {
  ValueBuilder? _value;
  ValueBuilder get value => _value ?? ValueBuilder();
  set value(ValueBuilder? newValue) => _value = newValue;

  ValueBuilder._();
  factory ValueBuilder() = _$ValueBuilder;
}
```
```
53: set value(ValueBuilder value) {                               '_$ValueBuilder.value=' ('void Function(ValueBuilder)') isn't a valid
                                                                  override of 'ValueBuilder.value=' ('void Function(ValueBuilder?)').
                                                                  #invalid_override

86: _$result = _$v ?? new _$Value._(value: super.value?.build()); (warning) The receiver can't be null, so the null-aware operator '?.' is
                                                                  unnecessary. #invalid_null_aware_operator

91: super.value?.build();                                         (warning) The receiver can't be null, so the null-aware operator '?.' is
                                                                  unnecessary. #invalid_null_aware_operator
```

samlythemanly:
Fixes #1173.

Previously, getters and setters could only be explicitly defined. This made sense for pre-null-safety, since there wasn't any reason to create a getter/setter pair over a normal field. However, with null safety, there was no way to use custom builders that had nested builders.

This fixes the bug where nested builders couldn't be used in custom builders, as well as allows the definition of abstract getters and setters such that auto created nested builders can also be utilized without needing null checks (by defining a setter with a nullable parameter).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+# 8.4.0
+
+- Allow abstract getters and setters in custom builders (necessary to use autocreated nested builders in custom builders).
+
 # 8.3.3
 
 - Fix erroneously generated null check for fields with generic bounds.
--- a/benchmark/pubspec.yaml
+++ b/benchmark/pubspec.yaml
@@ -1,5 +1,5 @@
 name: benchmark
-version: 8.3.3
+version: 8.4.0
 publish_to: none
 description: >
   Benchmark, not for publishing.
@@ -10,11 +10,11 @@ environment:
 
 dependencies:
   built_collection: ^5.0.0
-  built_value: ^8.0.0
+  built_value: ^8.4.0
 
 dev_dependencies:
   build_runner: '>=1.0.0 <3.0.0'
-  built_value_generator: ^8.3.3
+  built_value_generator: ^8.4.0
   pedantic: ^1.4.0
   quiver: '>=0.21.0 <4.0.0'
   test: ^1.0.0
--- a/built_value/pubspec.yaml
+++ b/built_value/pubspec.yaml
@@ -1,5 +1,5 @@
 name: built_value
-version: 8.3.3
+version: 8.4.0
 description: >
   Value types with builders, Dart classes as enums, and serialization.
   This library is the runtime dependency.
--- a/built_value_analyzer_plugin/pubspec.yaml
+++ b/built_value_analyzer_plugin/pubspec.yaml
@@ -1,5 +1,5 @@
 name: built_value_analyzer_plugin
-version: 8.3.3
+version: 8.4.0
 description: >
   Experimental analyzer plugin for the built_value code generator.
 homepage: https://github.com/google/built_value.dart
--- a/built_value_generator/lib/src/value_source_class.dart
+++ b/built_value_generator/lib/src/value_source_class.dart
@@ -889,83 +889,84 @@ abstract class ValueSourceClass
     result.writeln('$implName$_generics$orNull _\$v;');
     result.writeln('');
 
-    if (hasBuilder) {
-      for (var field in fields) {
-        final type = field.typeInCompilationUnit(compilationUnit);
-        final typeInBuilder = field.builderElementTypeWithPrefix;
-        final name = field.name;
+    for (var field in fields) {
+      late final String type;
 
-        var maybeOrNull = field.builderElementTypeIsNullable ? orNull : '';
-        if (field.isNestedBuilder) {
-          result.writeln('@override');
-          result.writeln('$typeInBuilder get $name {'
-              '_\$this;');
-          if (field.isAutoCreateNestedBuilder &&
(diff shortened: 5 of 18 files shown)


davidmorgan:
Thanks! This looks pretty good; I like that the setter/getter pair means you can mark that the builder will be auto-created.

I wonder if it would also make sense to support just having the builder as a nullable field? In the issue you commented that you get

```
Please make the following changes to use BuiltValue:

1. Make builder field value have type: Value? (or, if applicable, builder)
```

but I think--at least in my very quick checking--it works to just relax that check, so that if the field is nullable then it allows the nullable nested builder.

If both work then we could allow both, what do you think?

Thanks :)

davidmorgan:
Hmm, I'll merge this and try my suggestion on top. Thanks! Should be able to release in the next few days.

Issue #1198, opened 2022-12-03
Built Class with too many fields causes code generation to fail due to too much nesting in generated hashCode


knaeckeKami:
built_value_generator fails when trying to generate code for a class that has too many (several hundred) fields.

Repro: https://gist.github.com/knaeckeKami/4ea80b2c078abcf8717a6aa59b3aea9b

The output of running build_runner contains this error message:

```
[SEVERE] built_value_generator:built_value on lib/built_value_limits.dart:
An error `FormatterException` occurred while formatting the generated source for
  `package:built_value_limits/built_value_limits.dart`
which was output to
  `lib/built_value_limits.built_value.g.part`.
This may indicate an issue in the generator, the input source code, or in the
source formatter.
Could not format because the source could not be parsed:

line 3003, column 2001 of .: The file has too many nested expressions or statements.
     ╷
3003 │ $jc($jc($jc($jc($jc($jc($jc($jc($jc($jc($jc($jc($jc($jc($jc($jc($jc($jc($jc($jc($jc($jc($jc($jc($jc($jc($jc($jc($jc($jc($jc($jc($jc($jc($jc($jc($jc($jc($jc($jc($jc($jc($jc($jc($jc($jc($jc($jc($jc($jc($jc($jc($jc($jc($jc [...]
```

The issue is that the generated `hashCode` method combines the `hashCode` of all fields in a single nested expression and, when there are too many fields, this seems to break the dart formatter.

Also, I tested with built classes that have many fields (like 100), but not enough to break the dart formatter;
It seems that even these classes cause the dart formatter to slow down a lot, and this nested hashcode expression seems to play a role in that.

I created an example project with one single built_value class with 100 fields (similar to the repo above, but with only 100 instead of 1000 fields so that code generation still works) and measured the performance of `dart format .` with nested hashCode and with a version that does not nest the $jc() calls. 

Here are the outputs:

Nested version (current implementation of built_value_generator)

```
➜  dart format .
Formatted 4 files (0 changed) in 3.43 seconds.
➜  dart format .
Formatted 4 files (0 changed) in 3.39 seconds.
➜  dart format .
Formatted 4 files (0 changed) in 3.48 seconds.
```

flat version: 

```
dart format .
Formatted 4 files (0 changed) in 2.74 seconds.
➜  dart format .
Formatted 4 files (0 changed) in 2.60 seconds.
➜  dart format.
Formatted 4 files (0 changed) in 2.67 seconds.
```

[...]

PR #1199, opened 2022-12-03
refactor(built_value_generator): generate hashCode by using n statements for n fields instead of one nested expression that is nested n levels deep

knaeckeKami:
this should fix issue #1198.

### DIFF
--- a/built_value_generator/lib/src/value_source_class.dart
+++ b/built_value_generator/lib/src/value_source_class.dart
@@ -1209,7 +1209,7 @@ abstract class ValueSourceClass
       // Use a different seed for builders than for values, so they do not have
       // identical hashCodes if the values are identical.
       final seed = forBuilder ? 1 : 0;
-      result.writeln('int _\$hash  = $seed;');
+      result.writeln('var _\$hash  = $seed;');
 
       for (var field in comparedFields) {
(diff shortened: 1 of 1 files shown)


built_value_generator/lib/src/value_source_class.dart
```diff
@@ -1202,15 +1202,26 @@ abstract class ValueSourceClass
...
+      }
+
       // Use a different seed for builders than for values, so they do not have
       // identical hashCodes if the values are identical.
-      result.writeln(forBuilder ? '1, ' : '0, ');
-      result.write(
-          comparedFields.map((field) => '${field.name}.hashCode').join('), '));
-      result.writeln('));');
+      final seed = forBuilder ? 1 : 0;
+      result.writeln('int hash  = $seed;');
```
davidmorgan, line 1212: Please change to

`var _\$hash =`

--with `var` since there is no need to specify the type, and the `_\$` prefix to make it unlikely to clash with a field name, like we do for `_\$result` above.

knaeckeKami:
Thanks! 

Yes, I saw Object.hash and thought about using that, but I also feel like allocating an array for hash code collection is wasteful, especially in generated code where each performance sin is multiplied potentially a lot!

Good to see that it will be optimized in the future, though.

I applied your suggestions.

PR #1200, opened 2022-12-20
Drop dependency on quiver

natebosch:
The package had been used for only two utilities, `concat` and `zip`.

Replace `zip` with `IterableZip` from `package:collection` which is also
an existing dependency.

Replace `concat` with either `...` elements in a list literal (since
the collections are always iterated anyway) or by replacing `.map` with
`.expand`.


davidmorgan:
Thanks Nate! Makes sense.

Any particular urgency to landing / releasing / rolling to google3?

Thanks.

natebosch:
> Any particular urgency to landing / releasing / rolling to google3?

No urgency - I was just curious why I saw `quiver` in a pub get and saw that there was pretty little surface area used.

davidmorgan:
SG. Will land in the fullness of time ;) actually I have another change I should make, maybe I can do them together.

Thanks.

PR #1202, opened 2023-01-05
Add `non_constant_identifier_names` to ignore rule set

dnys1:
In certain cases, a `non_constant_identifier_name` lint can be triggered by the generator.

For example, consider the following declaration:

```dart
@SerializersFor([
  ...
])
final Serializers _serializers = _$_serializers;
```

The `_serializers` variable is meant to be private to the built_value type's file, however the resultant `_$_serializers` variable fails the `non_constant_identifier_names` lint causing `pana` errors when publishing.


davidmorgan:
Thanks! I recently learned that there is a way to tell the analyzer to ignore all lints, with "// ignore_for_file: type=lint", so I have merged https://github.com/google/built_value.dart/pull/1207 which does that. And we can finally stop playing catchup with the linter :)

So, I'll close this PR; please open another or an issue if there is anything not addressed by my PR.

For this specific lint, I am a little surprised that you want `serializers` to be private; it's intended that you have one `serializers` instance for the entire project, so that nested fields work. So wanting it to be private should be rare.

Thanks.

dnys1:
Ah great find! That should make codegen a lot easier 😄

The private serializers is definitely an odd pattern and one that I exclusively used in a few files to limit serialization to a few types in an attempt to reduce file size (since they were being compiled to JS). I didn't want to clutter the global namespace with a bunch of `serializers` choices so I went with the private name.

davidmorgan:
Makes sense, thanks :)

Issue #1203, opened 2023-01-06
Failed to precompile build script .dart_tool/build/entrypoint/build.dart


celvinren:
flutter pub run build_runner build --delete-conflicting-outputs
I ran the above cmd to generate .g file 1 months ago, but now it reporting this error message.
Try to do flutter clean and pub get again, but it doesn't resolve the issue.

Can anyone help me to fix this issue?

Thanks

`[INFO] Generating build script completed, took 351ms
[WARNING] /Users/calvin/.pub-cache/hosted/pub.dartlang.org/built_value_generator-8.4.2/lib/src/serializer_source_library.dart:125:43: Error: The getter 'element' isn't defined for the class 'DartType'.
 - 'DartType' is from 'package:analyzer/dart/element/type.dart' ('/Users/calvin/.pub-cache/hosted/pub.dartlang.org/analyzer-5.1.0/lib/dart/element/type.dart').
Try correcting the name to the name of an existing getter, or defining a getter or field named 'element'.
              SerializerSourceClass(type!.element as InterfaceElement)));
                                          ^^^^^^^
/Users/calvin/.pub-cache/hosted/pub.dartlang.org/built_value_generator-8.4.2/lib/src/dart_types.dart:26:31: Error: The getter 'element' isn't defined for the class 'DartType'.
 - 'DartType' is from 'package:analyzer/dart/element/type.dart' ('/Users/calvin/.pub-cache/hosted/pub.dartlang.org/analyzer-5.1.0/lib/dart/element/type.dart').
Try correcting the name to the name of an existing getter, or defining a getter or field named 'element'.
        ValueSourceClass(type.element as ClassElement).settings.instantiable;
                              ^^^^^^^
/Users/calvin/.pub-cache/hosted/pub.dartlang.org/built_value_generator-8.4.2/lib/src/dart_types.dart:31:12: Error: The getter 'element' isn't defined for the class 'InterfaceType'.
 - 'InterfaceType' is from 'package:analyzer/dart/element/type.dart' ('/Users/calvin/.pub-cache/hosted/pub.dartlang.org/analyzer-5.1.0/lib/dart/element/type.dart').
Try correcting the name to the name of an existing getter, or defining a getter or field named 'element'.
      type.element.allSupertypes
           ^^^^^^^
[...]

PR #1219, opened 2023-02-28
upgrade analyzer dependency to 5.2.0

nicholasphair:
[The `.element` interface was introduced (and `.element2` was deprecated)][1] in the analyzer at version 5.2.0.

Since, migrating over to the `.element` interface at 05da0d3d78, versions [less than 5.2.0 of the analyzer][2] will not work. 

Here, we update the analyzer dependency accordingly.

Resolves, #1203

[1]: https://github.com/dart-lang/sdk/commit/683e2419da03b68074c9c744d9c2566f51c588f5
[2]: https://github.com/google/built_value.dart/blob/v8.4.3/built_value_generator/pubspec.yaml#L12

### DIFF
--- a/built_value_generator/pubspec.yaml
+++ b/built_value_generator/pubspec.yaml
@@ -9,7 +9,7 @@ environment:
   sdk: '>=2.14.0 <3.0.0'
 
 dependencies:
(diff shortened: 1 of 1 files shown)


davidmorgan:
Thanks!

PR #1232, opened 2023-04-18
ISO 8601 duration serializer supporting microseconds

solwey-wes:
Added a new serializer with support for sub-second durations in ISO 8601 strings.  Eg: "PT9.977S".  I think this regex is solid and backwards compatible but worth a double-check...my regex game was pretty weak at the start of this task and GPT4 wasn't very helpful :P

I didn't try to add support for the longer units of time (weeks/months/years) since they weren't relevant to my use case.


davidmorgan:
Thanks.

I think we would only merge this if there is sufficient interest, i.e. if others are asking for it too. Otherwise it's simpler if you keep it as a local change. Thanks!

davidmorgan:
Closing as there has been no activity for years, please re-open if you'd like to keep going on this. Thanks.

PR #1234, opened 2023-04-19
escape field name on _$failedField

### LINKED ISSUE #1233: escaping invalid field names with a '$' breaks the generator on custom types
Our current problem:
We are building a builder that needs to generate custom Objects that can contain arbitrary field names and Types. We would like to escape invalid dart field names ba preceeding them with a `$`.
Like:
assert => $assert
class => $class
10 => $10

while this works when the field is one if Darts built in types this fails when we have custom types:
```dart
String get $assert;
int get $class;
DateTime get $10;
```
all work fine while:
```dart
CustomWhatever get $assert;
Custom get get $class;
Custom3 get $10;
```
all fail.

This can not be fixed by any suffix/prefix to the field name when being a number and only by a suffix on other reserved words:
```dart
CustomClass get johnDoeLoremIpsum$assert; //fails
CustomClass get johnDoe$10LoremIpsum // fails
CustomClass get $assertJohnDoe; // works
```
---

my guess is that the `$` isn't properly escaped somewhere (when creating the builder) thus it works for primitive dart types but doesn't for custom ones.

This means number fields without a letter are not possible :(

---

I've made a little POC for you to check out also appended below
https://github.com/Leptopoda/built_value_bug

```dart
abstract class BuggyClass implements Built<BuggyClass, BuggyClassBuilder> {
  const BuggyClass._();

  factory BuggyClass([
    void Function(BuggyClassBuilder)? b,
  ]) = _$BuggyClass;

  AnotherBuggyClass get $10;
}

abstract class AnotherBuggyClass
    implements Built<AnotherBuggyClass, AnotherBuggyClassBuilder> {
  const AnotherBuggyClass._();

  factory AnotherBuggyClass([
    void Function(AnotherBuggyClassBuilder)? b,
  ]) = _$AnotherBuggyClass;

  int get value1;
  int get value2;
}
```

fails generation with

```sh
➜  built_value_bug git:(main) ✗ dart pub run build_runner build
Deprecated. Use `dart run` instead.
[INFO] Generating build script completed, took 227ms
[INFO] Reading cached asset graph completed, took 37ms
[INFO] Checking for updates since last build completed, took 376ms
[SEVERE] built_value_generator:built_value on lib/built_value_bug.dart:
An error `FormatterException` occurred while formatting the generatesource for
  `package:built_value_bug/built_value_bug.dart`
which was output to
  `lib/built_value_bug.built_value.g.part`.
This may indicate an issue in the generator, the input source code, or in the
source formatter.
Could not format because the source could not be parsed:

line 90, column 19 of .: Expected an identifier.
   ╷
90 │ _$failedField = '$10'; $10.build();} catch (e) {
   │                   ^
   ╵
```

Leptopoda:
fixes #1233

Gonna sign the cla shortly.

### DIFF
--- a/end_to_end_test/lib/values_nnbd.dart
+++ b/end_to_end_test/lib/values_nnbd.dart
@@ -882,6 +882,10 @@ abstract class $ValueSpecial
   String? get aString;
   bool? get $mustBeEscaped;
 
+  SimpleValue? get $mustAlsoEscaped;
+  SimpleValue? get $assert;
+  SimpleValue? get $10;
+
   factory $ValueSpecial([void Function($ValueSpecialBuilder) updates]) =
       _$$ValueSpecial;
   $ValueSpecial._();
--- a/AUTHORS
+++ b/AUTHORS
@@ -11,3 +11,4 @@ Resul Alkan/MetGlobal <me@resulalkan.com>
 Dao Hoang Son <dao@hoangson.vn>
 Ahmed Fwela/Bdaya Development <ahmednfwela@bdaya-dev.com>
 Max Härtwig/Google Inc. <maxhaertwig@google.com>
+Nikolas Rimikis <rimikis.nikolas@gmail.com>
--- a/built_value_generator/lib/src/value_source_class.dart
+++ b/built_value_generator/lib/src/value_source_class.dart
@@ -1142,7 +1142,7 @@ abstract class ValueSourceClass
       result.write(fieldBuilders.keys.map((field) {
         final fieldBuilder = fieldBuilders[field];
         if (fieldBuilder == field) return '';
-        return "_\$failedField = '$field'; $fieldBuilder;";
+        return "_\$failedField = '${escapeString(field)}'; $fieldBuilder;";
       }).join('\n'));
 
       result.writeln('} catch (e) {');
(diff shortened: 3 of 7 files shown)


Leptopoda:
and signed :)

davidmorgan:
Thanks! Could you please add to the examples under the `end_to_end_test` package so this is covered? That will also give coverage of any related issues.

I suggest adding field names with $ to `values_nnbd.dart`, and covering serialization by having the types declare a serializer and adding to `serializers_nnb.dart`.

You can run `tool/presubmit` to regenerate all the generated files and run tests.

Thanks :)

Leptopoda:
done.
I thought it would fit nicely into `$ValueSpecial` and also added the serialization as requested.

end_to_end_test/lib/values_nnbd.dart
```diff
@@ -7,6 +7,7 @@ import 'package:built_collection/built_collection.dart';
 import 'package:built_value/built_value.dart';
 import 'package:built_value/serializer.dart';
 import 'package:end_to_end_test/enums_nnbd.dart' as using_import_as;
+import 'package:end_to_end_test/values.dart';
```
davidmorgan, line 10: It should work to remove this import, I think?

The presubmit is complaining because it's a "null safe" -> "legacy" import which isn't allowed.

Leptopoda: Oh my IDE must have added this while I was poking around without noticing :sweat_smile:

And I was wondering why the tool gave me such a 'complex' exit code (at least it was green).
Sorry for not noticing :)

davidmorgan:
Looks good now, thanks!

I'll aim to cut a release this week.

PR #1244, opened 2023-05-19
Fix codegen with records. Support record fields in value types.

davidmorgan:


built_value/pubspec.yaml
```diff
@@ -8,7 +8,7 @@ topics:
  - built-value
 
 environment:
-  sdk: '>=2.12.0 <3.0.0'
+  sdk: '>=2.12.0 <4.0.0'
```
ahmednfwela, line 11: since `RecordType` is 3.0.0 , will this cause problems for packages that are <3.0.0 ?

end_to_end_test/lib/records_nnbd.dart
```diff
@@ -0,0 +1,67 @@
...
+  // () get record0;
+  // ()? get record0;
+  //
+  // These need extra work to generate the "," as needed.
+  //
+  // (int,) get record1;
+  // (int,)? get record1n;
+
+  (int, int) get record2;
+  (int, int)? get record2n;
```
ahmednfwela, line 33: is it possible to also add tests for nullable values? e.g. (int?, int)

iinozemtsev (approved):
LG

PR #1273, opened 2023-10-31
Add explicit error message to enum ArgumentError

nhannah:
Presently when an unexpected value is provided for an enum without a fallback an `ArgumentError` is thrown containing only the unexpected value as the message. This provides little information to the developer and also prevents any meaningful introspection to be performed. I propose an error similar to [json_serializable](https://github.com/google/json_serializable.dart/blob/f7cb51662b21d11b3fdb6c5e0fdd46431cec59d4/json_annotation/lib/src/enum_helpers.dart#L38) to provide more clarity.

### DIFF
--- a/built_value_generator/lib/src/enum_source_class.dart
+++ b/built_value_generator/lib/src/enum_source_class.dart
@@ -202,7 +202,11 @@ abstract class EnumSourceClass
 
     var fallback = fields.firstWhereOrNull((field) => field.settings.fallback);
     if (fallback == null) {
-      result.writeln('default: throw new ArgumentError(name);');
+      result
+        ..writeln('default: throw new ArgumentError(')
+        ..writeln("'`\$name` is not one of the supported values: '")
+        ..writeln("'${fields.map((field) => field.name).join(', ')}',")
+        ..writeln(');');
     } else {
       result.writeln('default: return ${fallback.generatedIdentifier};');
     }
--- a/built_value_generator/test/enum_class_generator_test.dart
+++ b/built_value_generator/test/enum_class_generator_test.dart
@@ -45,7 +45,10 @@ TestEnum _$valueOf(String name) {
     case 'maybe':
       return _$maybe;
     default:
-      throw new ArgumentError(name);
+      throw new ArgumentError(
+        '`$name` is not one of the supported values: '
+        'yes, no, maybe',
+      );
   }
 }
 
@@ -482,7 +485,10 @@ TestEnum _$valueOf(String name) {
     case 'maybe':
       return _$yes;
     default:
-      throw new ArgumentError(name);
+      throw new ArgumentError(
+        '`$name` is not one of the supported values: '
+        'yes, no, maybe',
+      );
   }
 }
 
@@ -526,7 +532,10 @@ TestEnum _$vlOf(String name) {
     case 'maybe':
       return _$maybe;
     default:
-      throw new ArgumentError(name);
+      throw new ArgumentError(
+        '`$name` is not one of the supported values: '
+        'yes, no, maybe',
+      );
   }
 }
 


davidmorgan:
Thanks Neil.

I think I prefer the current message: it's true that it has less information than we _could_ provide, but I usually consider an exception message in combination with what you can get from the stack trace. The stack trace will take you straight to the EnumClass declaration--which will tell you what the valid values were.

Happy to be persuaded if enough people want it, though :) thanks.

nhannah:
Hey David,

The reason we want to add this is to allow for some level of runtime differentiation on Enum errors, even if it is via less than desirable introspection on the `ArgumentError` `message`. The reason we want to do this is because in the past we have put a lot of effort, coordination, and code into versioning REST endpoints when a new Enum value is required. Recently we reassessed that decision and realized that the vast majority of the time a user would need to update information on an updated client(or web), then re-access their account on an outdated client to possibly receive a "new" enum value on an old client that is not yet deprecated. This is an edge case scenario that we have decided we'd rather deal with by instructing a user their account is using features not supported by the client and they should update the client (in the past we either coerced new values to old or truncated responses of data not fitting the api contract). To do this in REST we have introspected on the `CheckedFromJsonException` to present a specific update client error screen if an enum deserialization exception is raised.

We have now recently started using graphQL with the Ferry client for Flutter which uses built_value for code generation. We have a Enum versioning solution on our backend, but it once again uses data coercion to not send "new" Enum values to old clients. It also requires extra code and is mostly focused on edge cases. We would like to remove the need to deal with these edge cases on the backend in the majority of cases by utilizing our same error introspection to direct edge case user exceptions to update their client. As it sits now we cannot differentiate Enum errors at runtime bubbled by Ferry though as they are just `ArgumentError`s with no differentiable message. This is why we would like to have a more verbose message added, to let us present the user with a more specific error in this Enum exception case.

davidmorgan:
Thanks for the extra detail :)

Deserialization is quite a different case from just calling EnumClass.valueOf by hand ... in the deserialization case you are handling external data, so you do need to handle failure somehow.

Doesn't it already get caught and rethrown as a `DeserializationError`? That should happen here:

https://github.com/google/built_value.dart/blob/master/built_value/lib/src/built_json_serializers.dart#L154C17-L154C37

Or possibly you are not using this codepath (`Serializers.deserialize`)?

nhannah:
Yep the `ArgumentError` gets bubbled up in a deeply nested `DeserializationError` within a `TypedLinkException` using Ferry/gql. I can locate the root error with some recursion, but the Enum `ArgumentError` cannot be differentiated from other `ArgumentError`s at that point so we can't treat it as "special". In all other cases we show a default error handling screen, only for an unhandled Enum case do we want to assume a versioning issue and direct users to update their client as the state that causes that case is predictable. 

```
extension on DeserializationError {
  bool hasNestedEnumArgumentError() {
    if (error is DeserializationError) return (error as DeserializationError).hasNestedEnumArgumentError();
    return error is ArgumentError &&
        (error as ArgumentError).message is String &&
        ((error as ArgumentError).message as String).contains('is not one of the supported values:');
  }
}
```

davidmorgan:
Got it.

An alternative route, probably not helpful now but worth mentioning in this context: you can mark one enum constant `@BuiltValueEnumConst(fallback: true)` and it will be returned for unrecognized values, instead of throwing.

From the `DeserializationError` you have the `FullType` of the type that failed to deserialize, so you have the _name_ of the enum--I don't suppose you have a list of all your enums somewhere? Then that would do it :)

You can also use the `FullType` to look up the serializer for the enum, but there is nothing special about enum serializers that would allow you to know they are enum serializers. Except ... they are probably the only generated `PrimitiveSerializer` s you are using. So you could do this: look up the serializer, check if it's a `PrimitiveSerializer`, if so check if it's one of the default serializers (instantiate a fresh Serializers, check if it's installed), if not, assum it's an enum serializer :)

Possibly we can come up with a better place to put this information so it's not so messy :)

nhannah:
> I don't suppose you have a list of all your enums somewhere

With graphQL / codegen these are all generated from the gql schema via CI, so while I can see them the `FullType` is not exactly clear or in my control, e.g. `GDateRangeOption`, `GUserTaskIdentifier`, and these could be added at will by the backend.

I can take a look at the `PrimitiveSerializer` approach later today.

davidmorgan:
Closing as there has been no activity for a year, please re-open if you'd like to keep going on this. Thanks.

Issue #1295, opened 2024-01-01
Record list serialization unexpectedly converted to a map by StandardJsonPlugin.


Leptopoda:
Hi there, 
I hope y'all had calm holidays and a good start into 2024.

### Problem:

I'm trying to serialize a record with a list into a json list. Unfortunately the `StandardJsonPlugin` is converting this record into a json map.

I have the following scenario:
```dart
typedef MyRecord = ({
  BuiltList<String>? list,
  int? number,
});

extension $MyRecordExtension on MyRecord {
  static Serializer<MyRecord> get _serializer => const MyRecordSerializer();
  static MyRecord fromJson(Object? json) => _$jsonSerializers.deserializeWith(_serializer, json)!;
  Object? toJson() => _$jsonSerializers.serializeWith(_serializer, this);
}

  test('MyRecord serialization', () {
    MyRecord object = (
      list: BuiltList<String>(),
      number: null,
    );

    Object? json = [];

    expect(object.toJson(), equals(json)); // results in {}
    expect($MyRecordExtension.fromJson(json), equals(object));

    object = (
      list: BuiltList<String>(['value1', 'value2', 'value3']),
      number: null,
    );

    json = ['value1', 'value2', 'value3'];

    expect(object.toJson(), equals(json)); // results in {"value1", "value2"}
    expect($MyRecordExtension.fromJson(json), equals(object));
  });
```

The registered serializer looks like this:
```dart
class MyRecordSerializer implements PrimitiveSerializer<MyRecord> {
  const MyRecordSerializer();

  @override
  Iterable<Type> get types => const [MyRecord];

  @override
  String get wireName => 'MyRecord';

  @override
  Object serialize(
    Serializers serializers,
    MyRecord object, {
    FullType specifiedType = FullType.unspecified,
  }) {
    dynamic value;
    value = object.list;
    if (value != null) {
      return serializers.serialize(value, specifiedType: const FullType(BuiltList, [FullType(String)]))!;
    }
    value = object.number;
    if (value != null) {
      return serializers.serialize(value, specifiedType: const FullType(int))!;
    }

	// Should not be possible after validation.
    throw StateError('Tried to serialize without any value.');
  }

  @override
  MyRecord deserialize(
    Serializers serializers,
    Object data, {
    FullType specifiedType = FullType.unspecified,
  }) {
    BuiltList<String>? list;
    try {
      list = serializers.deserialize(
        data,
[...]

PR #1296, opened 2024-01-09
feat: support ignoring custom types that StandardJsonPlugin should le…

Leptopoda:
…ave as lists

fixes: #1295

I was not sure how to best unit test this so I just used my own usecase.
Please suggest any other ways to test this if you don't like this approach.


davidmorgan:
Looks good, thanks!

Issue #1304, opened 2024-03-29
optimize building generated files


dave-doty:
Sorry, this may not be appropriate as a GitHub issue, but I just wanted to ask for advice/clarification on how `built_value` generates the `.g.dart` files.

My project has gotten to the point of using hundreds of `built_value` instances. It now takes about a minute to compile, sometimes two, which slows down development when testing several incremental changes consecutively.

In this comment with the OverReact authors: https://github.com/Workiva/over_react/issues/434#issuecomment-662475785, they suggest that built_value takes a while to compile on large projects because of its use of this package: https://pub.dev/packages/source_gen

I don't really understand the issues. I know OverReact itself generates code in `.g.dart` files, and I have about as many OverReact classes, of similar complexity, as built_value classes. Of course I don't know for sure that they are not also slowing things down, but according to the authors, their code generation is much faster.

So I'm wondering 

1. Whether you agree that source code generation is slower with `source_gen` than other methods.
2. If you think there is some way (whether by switching away from `source_gen`, or something else) to speed up compilation.

Thank you.

davidmorgan:
Thanks for filing! I'm definitely interested in performance issues.

The first thing I would try, if it happens to be easy, is splitting up your code into more packages. If you can arrange that the built_value types depend on less code then that should help.

But of course that may not be possible.

It would be interesting if I can see a repo of your setup, I don't suppose your code is on GitHub? :) if not then maybe you could describe the structure and I could go from that, e.g. how many files+class built_value and non-built_value in how many packages, how you're running the codegen. Then maybe I can make something comparable to experiment with.

Thanks!

jakemac53:
> The reason I mention multiple packages is that splitting to multiple packages can help to improve dependencies.

Note that nowadays, the resolver is also optimized when you split across multiple packages, it  won't actually read all transitive imports, we compute and output a digest of the transitive imports of a given file next to it, and if available will only read that instead. This can be an extremely dramatic improvement. But, it only works for package dependencies.

dave-doty:
Thank you for the suggestions, hopefully some of these will work for us.

OverReact only updated to Dart 3 with null safety a few days ago, we've been waiting for that to try to tackle upgrading ourselves, but as you said, it's got to be a big and tedious project.

davidmorgan:
As it happens I have a lot of experience with null safety migrations :) ... the migration tool is very helpful, but large packages like yours can be tough.

[...]

PR #1307, opened 2024-04-02
Cache parsed library results.

davidmorgan:
Benchmarking for #1304 shows that a lot of time is spent getting "ParsedLibraryResults", i.e. parsing files to get syntax trees.

Currently the parsing can be done multiple times per file because the results are not kept. In particular if there are N `built_value` classes in one file then it will be parsed at least `N` times.

So, cache the results within one generation.

Add a benchmark to `presubmit` that creates a simple file with 2500 built_value classes and builds it. Before this change it takes about 1m22s, after it takes about 25s, i.e. it's about 3.4x faster.


iinozemtsev (approved):
LG

PR #1309, opened 2024-04-10
Deprecate `@nullable` annotation

oprypin:
Legacy Dart is already unsupported and the annotation already causes an error whenever built_value can detect it, *but* it's possible to put this annotation into a completely unrelated place and it will just stay there doing nothing. So the least we can do is to deprecate this annotation.


davidmorgan:
SGTM, thanks!

Any particular urgency to merging this / rolling into google3? I'm travelling this week.

oprypin:
No urgency. I just thought it's quick to do

PR #1314, opened 2024-05-06
Move from pkg:pedantic to pkg:dart_flutter_team_lints

kevmoo:


davidmorgan:
Thanks!

Hmmmm there are some lint errors.

Looking at the lint set, I'm not super convinced ... ones that look worth fixing:

avoid_function_literals_in_foreach_calls
directives_ordering
unnecessary_lambdas
unnecessary_parenthesis
use_super_parameters

Problematic:

avoid_catching_errors ... is just wrong, sometimes you have to catch errors and this pushes people to do the wrong thing
prefer_const_constructors ... is a flutterism
prefer_relative_imports ... consistency here would be good but I'm not aware of a style push?
avoid_dynamic_calls, return_of_invalid_type, for_in_of_invalid_type, map_type_not_assignable, inference_failure_on_untyped_parameter ... these all relate to strict inference and will force changes to the code, are we sure those are equivalent?
lines_longer_than_80_chars ... I seem to remember this has false positives, maybe it doesn't any more
use_is_even_rather_than_modulo ... does not seem worth it

I didn't check for lints that are _not_ enabled after the switch from pedantic.

Not really sure where to go with this, any suggestions? :)

kevmoo:
More that pedantic is DEAD and trying to move away from it.

kevmoo:
I think we'd have to bump the min SDK on build_value first.

2.12 is a LOW base!

Issue #1335, opened 2024-12-18
Formatting generated code is slow


davidmorgan:
In `end_to_end_test`:

```
> dart format --language-version=3.6 lib/values.g.dart
Formatted 1 file (0 changed) in 2.61 seconds.
```

This is with the current/deprecated "short" formatting style. The new formatter is much faster:

```
dart format --language-version=3.7 lib/values.g.dart
Formatted lib/values.g.dart
Formatted 1 file (1 changed) in 0.76 seconds.
```

Note that there is overhead from the command startup, so the speedup is bigger than 3x.

Increasing the line width can help but it's not pretty and the `// dart format width=1000000` trick is only possible from 3.7; I'd like to land something that helps existing users.

It looks like it's this clause in the output that causes the problems:

```
      _$result = _$v ??
          new _$CompoundValueNoNestingField._(
              simpleValue: BuiltValueNullFieldError.checkNotNull(
                  simpleValue, r'CompoundValueNoNestingField', 'simpleValue'),
              validatedValue: validatedValue,
              simpleValueWithNested: simpleValueWithNested.build(),
              validatedValueWithNested: _validatedValueWithNested?.build());

```

and it can be made fast by adding one trailing comma.

```
      _$result = _$v ??
          new _$CompoundValueNoNestingField._(
            simpleValue: BuiltValueNullFieldError.checkNotNull(
                simpleValue, r'CompoundValueNoNestingField', 'simpleValue'),
            validatedValue: validatedValue,
            simpleValueWithNested: simpleValueWithNested.build(),
            validatedValueWithNested: _validatedValueWithNested?.build(),
          );

> dart format --language-version=3.6 lib/values.g.dart
Formatted 1 file (0 changed) in 0.72 seconds.
```

Both versions are fast with "tall" style.

I have a suspicion I'll want to backport this change, so I'll land the absolute minimal fix, i.e. this one comma, first.

 A format benchmark also looks like a good thing to have.

PR #1336, opened 2024-12-18
Generate code that formats faster with "short" style.

davidmorgan:
For #1335 

Before:

```
dart format --language-version=3.6 */lib/*.g.dart
Formatted 20 files (0 changed) in 3.55 seconds.
```

After:

```
dart format --language-version=3.6 */lib/*.g.dart
Formatted 20 files (0 changed) in 1.01 seconds.
```

Note that the speedup is more than 3.5x as "dart format" has some startup time.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 # 8.9.3 (unreleased)
 
+- Generate code that formats faster with current "short" style. Both versions
+  are fast with the coming "tall" style.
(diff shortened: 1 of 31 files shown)


iinozemtsev (approved):
LG

iinozemtsev:
so the trick is to proactively add a trailing comma?

davidmorgan:
> so the trick is to proactively add a trailing comma?

Right, that switches to the less-indented format that's more like the new "tall" style. And also turns out to be faster in this case :)

PR #1340, opened 2024-12-30
Migrate to use new analyzer element model.

scheglov:


scheglov:
I just wanted to save my work for now somewhere.
This version has dependency overrides, and cannot be landed as is.
New versions of `analyzer` is required.

kevmoo:
Can we run w/ this? It's blocking `code_builder` from getting the latest

davidmorgan:
@scheglov is this ready to work without dependency overrides? If so please update, I'm happy to merge+publish. Thanks!

scheglov:
I believe that with the analyzer `7.2.0` published, the override for the analyzer is not necessary anymore.

However we made a decision that we don't want  to publish `package:build` and correspondingly `package:source_gen` with two versions of API: old and new. Instead we will land these changes into branches, roll into google3, and then replace with new APIs.

So, `package:built_value` should go through the same process. Land into a branch, roll into google3, and then switch / rename to new APIs when the time is right - when everything is migrated at least this way.

davidmorgan:
Thanks @scheglov.

So if I understood correctly:

 - there's no way to publish this until `build` and `source_gen` have new versions published--is there a plan for when that will be, please?
 - but, you will need me to roll it into google3 before then to actually unblock that change--happy to do this, how urgent is it?

Thanks :)

scheglov:
1. Yes, no need to publish until everything migrated and landed into google3. Specific time is unknown. It depends on migrating everything else. There was a good progress.
2. I will run presubmit in google3 first, if it is clean, we can either land the CL into google3; or because it is consider wrong to land directly what is supposed to be imported from outside, land this PR into a branch, and roll the branch into google3. I will let you know if this package will become the last one, then it will become urgent. But in general, we need to do it.

scheglov:
https://critique.corp.google.com/cl/721562736 looks green.
So, how do you want to handle this?
Land into a branch on GitHub, and import this branch into google3?
Or land the CL directly?

davidmorgan:
@scheglov Thanks. I will merge into main and check the internal tests still make sense, it can just be unreleasable until we're ready for the next update. I'll send you a CL for google3.

davidmorgan:
@scheglov rolled in as cl/721696907. Thanks!

Issue #1343, opened 2025-02-25
built_value_generator dependencies


davidmartos96:
What's the progress on supporting the latest dependencies of the generator package? In particular `source_gen`. Other code gen packages have started requiring v2 of `source_gen` . 

I'm seeing that the repo main commit is currently using dependency overrides to more up to date dependencies.

davidmorgan:
Released `8.9.4` from a branch, hopefully this does the trick, if not please let me know :) thanks!

PR #1346, opened 2025-03-05
Support analyzer v7

davidmartos96:
As a followup of #1343 I tried analyzer version 7 on this branch and everything seems to be working. But I also saw #1340 saying there were pending changes, so I'm not sure if this is mergeable.

I've taken the opportunity to also migrate the chat_example to use the `web` package instead of `dart:html` to avoid analysis errors.

@davidmorgan

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,12 @@
 # Changelog
 
+# 8.9.5
+
+- Allow `built_value_generator` to use `analyzer 7.0.0`.
+
 # 8.9.4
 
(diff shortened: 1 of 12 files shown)


davidmorgan:
Thanks :) indeed it needs releasing from the branch; I've done that. (Thanks for spotting my typos!)

Could you please resend the chat example fixes as a new PR against main? Those we can just merge in :) thanks.

davidmartos96:
Will do. Should I revert them in this PR? Or do we keep them for CI checks purposes?

davidmorgan:
This PR is then no longer needed, I merged the version bump and released it :)

PR #1362, opened 2025-04-26
Update codegen documentation in README

victoreronmosele:
I updated the Flutter codegen command in the README from the deprecated `flutter packages pub run build_runner build` to `dart run build_runner build`.


davidmorgan:
Thanks!

It looks like the doc is really out of date :) `pub run` hasn't been the right command for a long while, so the paragraph above needs updating too.

I think it should be `dart run build_runner build` then remove the paragraph about Flutter altogether, what do you think? :)

victoreronmosele:
Just made the update @davidmorgan

